PR #7, opened 2016-09-02
manual: hack Makefiles and cmake to work together

junghans:
What hack, but it seems to work!


CMakeLists.txt
```diff
@@ -48,6 +48,9 @@ endif(NOT DEFINED MAN)
 if (NOT DEFINED DATA)
   set(DATA "share/votca")
 endif(NOT DEFINED DATA)
+if (NOT DEFINED XTP_MANUAL)
+  set(XTP_MANUAL "manual")
+endif(NOT DEFINED XTP_MANUAL)
```
junghans: This is a really strange name! I would call it `MANUAL_DIR` or something.

manual/Makefile
```diff
@@ -44,8 +44,10 @@ endif
 qclean:
 	$(LATEXMK) -C $(NAME).tex
 
-install: all
-	@echo "$(NAME).pdf is ready to use, no need to install it."
+install: xtp_manual.pdf
+	mkdir -p $(XTP_MANUAL)
+	cp $< $(XTP_MANUAL)/xtp_manual.pdf
+	echo "Installing to ${XTP_MANUAL}"
```
junghans: That will break packaging, which uses the `DESTDIR` variable. Use
```
	mkdir -p $(DESTDIR)$(XTP_MANUAL)
	cp $< $(DESTDIR)$(XTP_MANUAL)/xtp_manual.pdf
```
and test with:
```
$ make manual
$ make installmanual DESTDIR=/tmp
```

PR #19, opened 2017-07-20
patch for xidft when tasks="input" (and not "project" or "extract")

JensWehner:


floaltvater:
Sorry, I actually have another version, which basically just checks that _jre_summary has the "output" attribute. And there are some other things going on as well. I just haven't gotten around to pushing it. That's why I hadn't submitted the pull request yet.

JensWehner:
ahh sorry

PR #36, opened 2017-08-04
Wall

JensWehner:


src/libxtp/gwbse/gwbseengine.cc
```diff
@@ -148,9 +148,9 @@ namespace votca {
             // parse DFT data, if required
             if (_do_dft_parse) {
                 if (_redirect_logger) {
-                    CTP_LOG(ctp::logINFO, _gwbse_engine_logger) << "Parsing DFT data from " << _dftlog_file << " and " << _MO_file << flush;
+                    CTP_LOG_SAVE(ctp::logINFO, _gwbse_engine_logger) << "Parsing DFT data from " << _dftlog_file << " and " << _MO_file << flush;
```
JensWehner: I do not understand why this is necessary?

baumeier: For what it is worth, neither do I...

junghans: `CTP_LOG` is a macro defined in `include/votca/ctp/logger.h`:
```c
#define CTP_LOG(level, log) \
if ( &log != NULL && level > (log).getReportLevel() ) ; \
else (log)(level)
```
and in this function `log` is never `NULL` and hence the compiler complains about a useless comparison:
 ```
/home/votca/votca/xtp/src/libxtp/gwbse/gwbseengine.cc: In member function 'void votca::xtp::GWBSEENGINE::ExcitationEnergies(votca::xtp::QMPackage*, std::vector<votca::ctp::Segment*>, votca::xtp::Orbitals*)':
/home/votca/votca/ctp/include/votca/ctp/logger.h:35:11: warning: the address of '_gwbse_engine_logger' will never be NULL [-Waddress]
 if ( &log != NULL && level > (log).getReportLevel() ) ; \
           ^
/home/votca/votca/ctp/include/votca/ctp/logger.h:35:11: warning: the address of '_gwbse_engine_logger' will never be NULL [-Waddress]
```
(https://travis-ci.org/votca/votca/jobs/260781933#L5116)

junghans: On a related note, we should switch to `boost::logging`.

junghans:
Now all warnings are gone in votca/votca#19, so ready to merge.

PR #41, opened 2017-08-09
Orca aux

baumeier:
Orca gets a custom auxbasis writer to the input file. Also a bit of cleanup.

### DIFF
--- a/src/libxtp/qmpackages/orca.cc
+++ b/src/libxtp/qmpackages/orca.cc
@@ -310,7 +310,7 @@ namespace votca {
 
         /**
          * Prepares the *.inp file from a vector of segments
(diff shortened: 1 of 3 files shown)


src/libxtp/qmpackages/orca.cc
```diff
@@ -36,837 +36,862 @@
...
+                    if ((*it)->charge != 0.0) _crg_file << fmt << endl;
+                }
             }
+            return;
         }
-    } // write pseudopotentials
 
+        /**
+         * Prepares the *.inp file from a vector of segments
+         * Appends a guess constructed from monomer orbitals if supplied
```
JensWehner: Does it supply a guess?

PR #51, opened 2017-09-01
Svs2 v

JensWehner:
RI-V instead of SVS refactoring of QMpackages, qmpackages now reorder, everything else uses VOTCA ordering. This possibly breaks turbomole parser because no atoms are parsed in turbomole.cc . Removed cpmd as well.


src/libxtp/aomatrices/aocoulomb.cc
```diff
@@ -1143,52 +1143,39 @@ if (_lmax_col > 5) {
...
-            
-            _aomatrix = ub::prod(_aomatrix , _gwoverlap_inverse);
- 
-
-        }
-        else{
-            cout<<"WARNING: gwbasis is not stable."<<endl;
+    int AOCoulomb::Symmetrize(const ub::matrix<double>& _gwoverlap_cholesky){
+        
+       //This converts V into L(LT V L)-1/2 LT, which is needed to construct 4c integrals,
```
baumeier: Are you documenting this somewhere in more detail? We need that for eventual alternatives to PPM.

JensWehner: To be honest, I have no real clue, why this worked. The non orthogonal space has two representations connected by the S matrix. I think without the ppm you do not have to use this expression but V-1/2 might be enough.

PR #58, opened 2017-10-26
Othertry

JensWehner:
fixes to kmc 
speed up of gencube
added lothars cutoff for 3cc
aomatrices do not have to be initialized anymore


JensWehner:
merged it with master, what do you mean?

baumeier:
State indexing is messed up for the case when KS states taken into account do not start at zero.

PR #68, opened 2018-01-08
Gwbse output patch

zetadin:
The output property only exists if we GWBSE was enabled. Otherwise, exciton exits with an error message "property not found: output". This was causing failure in post-merge tests of my branch.


JensWehner:
makes senes. Agreed

zetadin:
Ok, then I'll merge. Pull your branch.

PR #76, opened 2018-02-09
.travis.yml: fix for docker network issues

junghans:


junghans:
The one when Travis times out!

PR #77, opened 2018-02-10
.travis.yml: timeout docker pull

junghans:


junghans:
That should have been part of #76

PR #80, opened 2018-02-15
Create test dir w cmake lists

JoshuaSBrown:
Basically setting up the directory and cmake files for testing using the boost framework. Used the glink.h file as a dummy test.


JoshuaSBrown:
That's what I do 👍.

PR #89, opened 2018-03-08
Segment migration

JoshuaSBrown:
Migration of segment and segmenttype classes to xtp. Namespace changes as well primitive testing.


JoshuaSBrown:
@junghans is this better?

src/libxtp/segment.cc
```diff
@@ -0,0 +1,542 @@
...
+  _CoM = _CoM + shift;
+
+  vector<Fragment *>::iterator fit;
+  for (fit = _fragments.begin(); fit < _fragments.end(); fit++) {
+
+    (*fit)->TranslateBy(shift);
+  }
+}
+
+void Segment::setHasState(bool yesno, int state) {
```
JensWehner: I do not like this accessor way, can we discuss using maps instead, which have hole, electron, gs, singlet, triplet as keys or (e,h,n,s,t)

src/libxtp/segment.cc
```diff
@@ -0,0 +1,542 @@
...
+    _U_cC_nN_h = dU;
+  } else {
+    throw runtime_error(" ERROR CODE whe__00u11a__");
+  }
+}
+
+void Segment::setU_nC_nN(double dU, int state) {
+
+  if (state == -1) {
+    _U_nC_nN_e = dU;
```
JensWehner: sort of same here. Using maps could greatly improve readibilities

src/libxtp/segment.cc
```diff
@@ -0,0 +1,542 @@
...
+    for (fit = this->Fragments().begin(); fit < this->Fragments().end();
+         fit++) {
+      (*fit)->Rigidify();
+    }
+  } else {
+    return;
+  }
+}
+
+void Segment::WritePDB(FILE *out, string tag1, string tag2) {
```
JensWehner: another pdb writer

src/libxtp/segment.cc
```diff
@@ -0,0 +1,542 @@
...
+              0.0,                   // Temperature factor           %6.2f
+              " ",                   // Segment identifier           %4s
+              name.c_str(),          // Element symbol               %2s
+              " ",                   // Charge on the atom.          %2s
+              (*pol)->getQ00());
+    }
+  }
+}
+
+void Segment::WriteXYZ(FILE *out, bool useQMPos) {
```
JensWehner: xyz writer

include/votca/xtp/segment.h
```diff
@@ -0,0 +1,199 @@
...
+#include <votca/tools/vec.h>
+
+namespace votca { namespace xtp {
+
+class Atom;
+class PolarSite;
+class APolarSite;
+class Fragment;
+class SegmentType;
+class Topology;
```
junghans: Can we use the topology from csg?

src/libxtp/fragment.cc
```diff
@@ -46,20 +46,22 @@ Fragment::Fragment(Fragment *stencil)
...
     _atoms.clear();
 
     vector < PolarSite* > ::iterator pit;
-    for (pit = this->PolarSites().begin();
+/*    for (pit = this->PolarSites().begin();
             pit < this->PolarSites().end();
             ++pit) {
         delete *pit;
-    }
+    }*/
```
junghans: Just remove it!

src/libxtp/segment.cc
junghans: Right!

src/libxtp/segment.cc
JoshuaSBrown: I don't want to change any of the methods at this stage. I would prefer to wait until I get the dependence of xtp on these classes switched and working and then refactor, but yes let's discuss. What is this method doing? This is the one that is describing if the state is charged and or excited?

src/libxtp/segment.cc
JoshuaSBrown: Yes completely agree. Can you give me an outline of what you are thinking?

include/votca/xtp/segment.h
JoshuaSBrown: I think it already does I just have not imported the topology class from ctp yet.

src/libxtp/segment.cc
JoshuaSBrown: I can consolidate this with the pdb reader and writer in csg if this is useful for this class to have them e.g. for testing or whatever. Or if it is not useful we can just get rid of it.

src/libxtp/segment.cc
JoshuaSBrown: Same with the xyz writer.

src/libxtp/fragment.cc
JoshuaSBrown: ok

src/libxtp/segment.cc
JensWehner: yes, but actually we should aim for a better separation of state and topology, I still do not fully get what is what

src/libxtp/segment.cc
junghans: Maybe get rid of them for now and relink to the csg readers when needed later.

JoshuaSBrown: Fair enough.

JensWehner: I personally think the file readers and writers should allow every atom container to write and read files.

src/tests/CMakeLists.txt
```diff
@@ -1,7 +1,7 @@
 if(ENABLE_TESTING)
     find_package(Boost 1.39.0 REQUIRED COMPONENTS unit_test_framework)
-    foreach(PROG test_glink test_atom test_fragment test_polarsite test_apolarsite)
-      file(GLOB ${PROG}_SOURCES ${PROG}*.cc)
+    foreach(PROG test_glink test_atom test_fragment test_segment test_segmenttype test_polarsite test_apolarsite)
+      file(GLOB ${PROG}_SOURCES ${PROG}.cc)
```
junghans: Either remove the glob or rename `test_segment’` to `test_segment_base`!

Globbing without a wildcard is just confusing!

PR #93, opened 2018-03-09
4centers onthefly

JavierSijen:
Implementation of direct 4c integrals.

DONE;
DFT

TODO:
DFT prescreen
options
GW?


src/libxtp/ERIs.cc
```diff
@@ -192,7 +192,116 @@ namespace votca {
...
+			cout << "ERIS.cc ERIs::CalculateERIs_4c_direct" << endl;
+			
+			// Initialize ERIs matrix
+			_ERIs = ub::zero_matrix<double>(DMAT.size1(), DMAT.size2());
+			// Get number of shells
+			int numShells = dftbasis.getNumofShells();
+			
+			// Test
+			printf("size(DMAT) = [%d, %d]\n", int(DMAT.size1()), int(DMAT.size2()));
+			printf("dftbasisSize = %d\n", dftbasis.AOBasisSize());
```
JoshuaSBrown: Any reason for using printf(C) vs cout(C++) here?

src/libxtp/ERIs.cc
```diff
@@ -192,7 +192,116 @@ namespace votca {
...
+			printf("numShells = %d\n", numShells);
+
+			// Timer
+			clock_t start = clock();
+
+			#pragma omp parallel for
+			for (int iShell_3 = 0; iShell_3 < numShells; iShell_3++)
+			{
+				const AOShell* shell_3 = dftbasis.getShell(iShell_3); // Shell object
```
JoshuaSBrown: There are several shared objects here have you run any calculations to see how efficient the parallelization is. If you have shared memory access between threads the performance can actually be worse than running it in serial.

JensWehner: Yes and no, 1) The shell objects are pretty large, so it is unlikely that false sharing happens. Secondly, the are read access only, which also reduces the risk. I agree that maybe we should rewrite AOBasis as a continguous vector without any pointers.

src/libxtp/ERIs.cc
JavierSijen: Thanks for the comment, Joshua!

I will work on the formatting. Also, I'm used to methods similar to printf from other languages. Guess it would make more sense to use cout in this case!

src/libxtp/dftengine/dftengine.cc
```diff
@@ -244,7 +244,8 @@ namespace votca {
                     if (_with_RI) {
                         _ERIs.CalculateERIs(_dftAOdmat);
                     } else {
-                        _ERIs.CalculateERIs_4c_small_molecule(_dftAOdmat);
+						//_ERIs.CalculateERIs_4c_small_molecule(_dftAOdmat);
+						_ERIs.CalculateERIs_4c_direct(_dftbasis, _dftAOdmat);
```
JensWehner: make it optional as planned

src/libxtp/ERIs.cc
```diff
@@ -192,7 +192,116 @@ namespace votca {
...
+					int start_4 = shell_4->getStartIndex();
+					int numFunc_4 = shell_4->getNumFunc();
+
+					for (int iShell_1 = 0; iShell_1 < numShells; iShell_1++)
+					{
+						const AOShell* shell_1 = dftbasis.getShell(iShell_1);
+						int start_1 = shell_1->getStartIndex();
+						int numFunc_1 = shell_1->getNumFunc();
+
+						for (int iShell_2 = 0; iShell_2 < numShells; iShell_2++)
```
JensWehner: you iterate over all combinations of i,j,k,l instead due to symmetry you should only iterate over 1/8 of it and then add the result to all respective combinations.

JoshuaSBrown:
Sorry I should have written this earlier but is there anyway we can put a test for this part of the code in the /src/tests folder? 
@JensWehner does this seem appropriate?

Also @JavierSijen don't know if you are aware but there is a developers guide here: https://github.com/votca/votca/blob/master/DEVELOPERS_GUIDE.md
if you are planning to implement more changes it might be worth looking through, though your code seems to be well written anyway.

JensWehner:
Please check the continuous integration and remove all the errors as well. Otherwise good job

JensWehner:
Yes there should be a test of it. After the eigen conversion.

src/libxtp/ERIs.cc
```diff
@@ -192,6 +196,238 @@ namespace votca {
...
+              } // End loop over shell 4
+            } // End loop over shell 3
+            
+            #pragma omp critical
+            { // Begin omp critical
+              
+              // Add contributions of the (1, 2) <--> (3, 4) symmetry to ERIs matrix
+              for (int i = 0; i < DMAT.size1(); i++)
+                for (int j = i; j < DMAT.size2(); j++)
+                  _ERIs(i, j) += ERIsSymm(i, j);
```
JensWehner: `_ERIs+=ERIsSymm`

src/libxtp/ERIs.cc
```diff
@@ -192,6 +196,238 @@ namespace votca {
...
+          for (int i = 0; i < DMAT.size1(); i++)
+            for (int j = i + 1; j < DMAT.size2(); j++)
+              _ERIs(j, i) = _ERIs(i, j);
+
+          CalculateEnergy(DMAT.data());
+          return;
+        }
+        
+        
+        void ERIs::FillERIsBlock(ub::matrix<double> &ERIsCur, const ub::matrix<double> &DMAT, const ub::matrix<double> &subMatrix, const AOShell *shell_1, const AOShell *shell_2, const AOShell *shell_3, const AOShell *shell_4) {
```
JensWehner: can you make the shells references and not pointers

src/libxtp/ERIs.cc
```diff
@@ -192,6 +196,238 @@ namespace votca {
...
+          _diagonals = ub::zero_matrix<double>(dftBasisSize);
+          
+          for (int iShell_1 = 0; iShell_1 < numShells; iShell_1++) {
+            const AOShell* shell_1 = dftbasis.getShell(iShell_1);
+            for (int iShell_2 = iShell_1; iShell_2 < numShells; iShell_2++) {
+              const AOShell* shell_2 = dftbasis.getShell(iShell_2);
+              
+              // Get the current 4c block
+              ub::matrix<double> subMatrix = ub::zero_matrix<double>(shell_1->getNumFunc() * shell_1->getNumFunc(), shell_2->getNumFunc() * shell_2->getNumFunc());
+              bool nonzero = _fourcenter.FillFourCenterRepBlock(subMatrix, shell_1, shell_1, shell_2, shell_2);
```
JensWehner: should it not be (12|12)

src/libxtp/ERIs.cc
```diff
@@ -192,6 +196,238 @@ namespace votca {
...
+                } // End loop over functions in shell 2
+              } // End loop over functions in shell 1
+            } // End loop over shell 2
+          } // End loop over shell 1
+        }
+        
+
+        bool ERIs::CheckScreen(double eps, const AOShell* shell_1, const AOShell* shell_2, const AOShell* shell_3, const AOShell* shell_4) {
+          
+          if (!_with_screening)
```
JensWehner: this check should be outside this method

src/libxtp/ERIs.cc
```diff
@@ -192,6 +196,238 @@ namespace votca {
...
+                for (int iFunc_2 = 0; iFunc_2 < shell_2->getNumFunc(); iFunc_2++) {
+                  int ind_2 = shell_2->getStartIndex() + iFunc_2;
+
+                  // Symmetry
+                  if (ind_1 > ind_2)
+                    continue;
+
+                  // Cauchy–Schwarz
+                  // <ab|cd> <= sqrt(<ab|ab>) * sqrt(<cd|cd>)
+                  double ub = std::sqrt(_diagonals(ind_1, ind_2) * _diagonals(ind_3, ind_4));
```
JensWehner: use <ab|ab>*<cd|cd>< eps^2 instead because sqrt are expensive

src/libxtp/dftengine/dftengine.cc
```diff
@@ -1061,5 +1070,19 @@ namespace votca {
 
             return avdmat;
         }
+
+        
+        
+        void DFTENGINE::CalculateERIs_4c(const AOBasis& dftbasis, const ub::matrix<double> &DMAT) {
+
+          if (_with_RI)
+            _ERIs.CalculateERIs(_dftAOdmat);
```
JensWehner: I like that you use more functions and make them smaller, I would prefer if the input is checked in initialize and not later, e.g. there is a method in tools : ` T ifExistsAndinListReturnElseThrowRuntimeError(const string &key,std::vector<T> possibleReturns); `, 

which can be called  e.g.
`std::vector<std::string> choices={"direct","cache","RI"};
`result=ifExistsAndinListReturnElseThrowRuntimeError(input ,choices); `

junghans:
Travis says:
```
/home/votca/votca/xtp/src/libxtp/ERIs.cc:299:29: error: comparison between signed and unsigned integer expressions [-Werror=sign-compare]
           for (int i = 0; i < DMAT.size1(); i++)
                           ~~^~~~~~~~~~~~~~
/home/votca/votca/xtp/src/libxtp/ERIs.cc:300:35: error: comparison between signed and unsigned integer expressions [-Werror=sign-compare]
             for (int j = i + 1; j < DMAT.size2(); j++)
                                 ~~^~~~~~~~~~~~~~
cc1plus: all warnings being treated as errors
```

JensWehner:
@JavierSijen please make sure travis builds your branch.

junghans:
@JensWehner: is this ready to be merged?

JensWehner:
@junghans We close this pull request and merge it directly into the eigen branch because unit tests are easier there

PR #97, opened 2018-03-24
Merge in ceres solver finding and linking

### LINKED ISSUE #96: Add ceres-solver detection to cmake
[ceres-solver](http://ceres-solver.org) contains a number of optimizers that we are interested in using.

pmadhikar:
Ceres-solver will be used for optimisation in xtp. The current version in the ubuntu 16.04 repos is missing a header file somehow, see [bug report](https://bugs.launchpad.net/ubuntu/+source/ceres-solver/+bug/1595692), so one should compile it manually for now. This should not be a problem in newer, and maybe older, debian based distros - only 16.04. 

fixes #96


src/libxtp/bfgs-trm.cc
```diff
@@ -123,7 +123,7 @@ namespace votca {
 
                 } // checking step to be trusted
 
-                // after the step is accepted, we can shift the stored data
```
pmadhikar: My editor is configured to automatically delete trailing spaces. I must''ve forgotten to turn that setting off at some point. These changes are not related to code.

CMakeLists.txt
```diff
@@ -92,6 +92,12 @@ else()
 	set(LIBXC_LIBRARIES)
 endif()
 
+find_package(CERES)
+if(CERES_FOUND)
+  message("-- Found ceres solver version ${CERES_VERSION}")
```
junghans: Use `message(STATUS ”...“)`

CMakeLists.txt
```diff
@@ -92,6 +92,12 @@ else()
...
+  include_directories(${CERES_INCLUDE_DIRS})
+endif(CERES_FOUND)
```
junghans: Add an `else` case with empty `CERES_LIBRARIES`

junghans:
@JensWehner for the `MINIMAL` build Ceres and eigen needs to be added to the disable block in the Dockerfile of votca/votca!

JensWehner:
Eigen and Ceres will be required elements of XTP very soon

junghans:
Ok, then we need to add `REQUIRED` to the `find_package()` call!

PR #98, opened 2018-03-26
better kmcoutput every 1e9 steps

JensWehner:
Output of diffusion tensor every 1e9 steps

### DIFF
--- a/src/libxtp/calculators/kmcmultiple.cc
+++ b/src/libxtp/calculators/kmcmultiple.cc
@@ -39,7 +39,7 @@ void KMCMultiple::Initialize(tools::Property *options){
     _seed=options->ifExistsReturnElseThrowRuntimeError<int>(key+".seed");
     _numberofcharges=options->ifExistsReturnElseThrowRuntimeError<int>(key+".numberofcharges");
     _injection_name=options->ifExistsReturnElseThrowRuntimeError<std::string>(key+".injectionpattern");
-  
+  _intermediateoutput_frequency=options->ifExistsReturnElseReturnDefault<int>(key+".intermediateoutput",1E9);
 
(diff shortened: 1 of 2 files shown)


src/libxtp/calculators/kmcmultiple.cc
```diff
@@ -284,6 +286,12 @@ void KMCMultiple::RunVSSM(ctp::Topology *top)
                 avgdiffusiontensor += (_carriers[i]->dr_travelled)|(_carriers[i]->dr_travelled);
             }
         }
+        
+        if(step!=0 && step%(int(1e9))==0){
```
baumeier: Why have 1e9 hard coded?

src/libxtp/calculators/kmcmultiple.cc
```diff
@@ -284,6 +286,12 @@ void KMCMultiple::RunVSSM(ctp::Topology *top)
...
+            unsigned long diffusionsteps=step/diffusionresolution;
+            tools::matrix result=avgdiffusiontensor/(diffusionsteps*2*simtime*_numberofcharges);
+            cout<<endl<<"Step: "<<step<<" Diffusion tensor averaged over all carriers (nm^2/s):" << endl << result << endl;
```
baumeier: Can we have the same for average velocity, useful in mobility calculation

src/libxtp/calculators/kmcmultiple.cc
JensWehner: yes, we can set it as a default option.

src/libxtp/calculators/kmcmultiple.cc
JensWehner: yes, okay

PR #108, opened 2018-04-11
Removed unnecessary elements.h file as it is in tools

JoshuaSBrown:
This file was never used in xtp and is also present in tools


JoshuaSBrown:
I don't think anything uses it.

JoshuaSBrown:
It builds but wait for the tests to finish.

junghans:
Would it be better to merge #109 and then merge master back into `development`.

JoshuaSBrown:
I'm not a git master... so I'll do whatever you think is a good idea, but I think the development branch is now quite different from the master in some ways so there will be extra stuff that needs to be done. (maybe)

JoshuaSBrown:
@JensWehner ya it didn't build...

junghans:
In a nutshell, if you have to do a commit twice, you are doing something wrong! Just add it one branch and merge it into the other. So here, add it to `master` and then merge `master` into `development`.

JensWehner:
Dismissed because of master merge

PR #111, opened 2018-04-13
Hdf5 checkpoint

### LINKED ISSUE #15: HDF5 orb file
What savefile should contain

# of external charges
bool APEorNot
number of alpha electrons
number of beta electrons
alpha/beta MOs reordert in xtp
alpha/beta MOEnergies
qm_atoms
TotalGSEnergy
Basisset name and object
ECP name and object
Vxc
Functional name
frac of EXX
Original qmpackage name
GW pert energies
GW diag energies and states
GW aux basisset name and object
BSE S/T Energies
BSE S/T Coefficients
BSE S/T AR_Coefficients
BSE Hamiltionian parts

maybe:
Fragment pops
Transition dipoles

pmadhikar:
Fixes #15 

### Missing things: 
- Documentation
- Thorough error checking 

Other than the above, it seems to be working well. Any future changes should, in theory, be limited to the checkpoint file library only. In other words, the interface to the outside is pretty okay I think.


include/votca/xtp/checkpoint.h
```diff
@@ -0,0 +1,324 @@
...
+ private:
+  std::string _fileName;
+  H5::H5File _fileHandle;
+  std::string _version;
+};
+
+
+}  // namespace xtp
+}  // namespace votca
+#endif  // CHECKPOINT_H
```
JoshuaSBrown: Is there any particular reason to include all three classes in the same file? The header file could benefit from some doxygen documentation.

include/votca/xtp/orbitals.h
```diff
@@ -1,4 +1,4 @@
-/* 
+/*
  *            Copyright 2009-2017 The VOTCA Development Team
```
JoshuaSBrown: 2009-2018

src/libxtp/checkpoint.cc
```diff
@@ -0,0 +1,71 @@
+/*
+ *            Copyright 2009-2017 The VOTCA Development Team
```
JoshuaSBrown: 2018

CMakeLists.txt
```diff
@@ -90,6 +90,20 @@ if(CERES_FOUND)
...
+if(HDF5_FOUND)
+  message("-- HDF5 found for checkpointing")
+  message("   Version " + ${HDF5_VERSION})
+  include_directories(${HDF5_INCLUDE_DIRS})
+
+  if (${HDF5_VERSION} GREATER 1.8)
+    message(WARNING "HDF5 will be used such that it is compatible with version 1.8.")
+  endif()
+
+endif()
```
junghans: What is happening if `HDF5` isn't found?

CMakeLists.txt
```diff
@@ -90,6 +90,20 @@ if(CERES_FOUND)
...
 
+find_package(HDF5 1.8 COMPONENTS CXX)
+
+if(HDF5_FOUND)
+  message("-- HDF5 found for checkpointing")
+  message("   Version " + ${HDF5_VERSION})
+  include_directories(${HDF5_INCLUDE_DIRS})
+
+  if (${HDF5_VERSION} GREATER 1.8)
+    message(WARNING "HDF5 will be used such that it is compatible with version 1.8.")
```
junghans: Isn't the check for version 1.8 already part of the above `find_package(HDF5 ..)` line?

include/votca/xtp/checkpoint.h
pmadhikar: Not really, it is mostly laziness on my part. I'll move them to their own files...

include/votca/xtp/checkpoint.h
```diff
@@ -0,0 +1,324 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ */
+
+#ifndef CHECKPOINT_H
```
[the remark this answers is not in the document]
pmadhikar: Aah, good catch.

CMakeLists.txt
pmadhikar: I think HDF5 will be required as we are getting rid of the old orb files. So CMake should fail if there isn't any HDF5.

CMakeLists.txt
pmadhikar: Indeed it is. This is because HDF5 versions greater than 1.8 have some api changes in them that are not compatible with 1.8. They can be set to work in "1.8 mode", which is the default behaviour for now, but might need to be changed in the future.

include/votca/xtp/orbitals.h
pmadhikar: Thanks again, I'm really bad at keeping the comments correct.

CMakeLists.txt
junghans: Ok, so you add `REQUIRED` above.

include/votca/xtp/checkpoint.h
```diff
@@ -0,0 +1,324 @@
...
+#include <Eigen/Eigen>
+#include <H5Cpp.h>
+#include <string>
+#include <typeinfo>
+#include <vector>
+#include <type_traits>
+#include <votca/tools/vec.h>
+
+
+namespace votca {
```
JensWehner: I like the code but some explanation would be useful of what it does

include/votca/xtp/checkpoint.h
```diff
@@ -0,0 +1,324 @@
...
+
+    // see the following links for details
+    // https://stackoverflow.com/a/8671617/1186564
+    template <typename T>
+        typename std::enable_if<!std::is_fundamental<T>::value>::type
+        operator()(const T& data, const std::string& name){
+        WriteData(_loc, data, name);
+    }
+
+    // Use this overload iff T is a fundamental type
```
JensWehner: typo

include/votca/xtp/checkpoint.h
```diff
@@ -0,0 +1,324 @@
...
+            CptLoc tempGr = loc.createGroup("/" + name);
+            WriteData(tempGr, x.second, "index" + r);
+            ++c;
+        }
+    }
+
+};
+
+
+class Reader{
```
JensWehner: separate file please

include/votca/xtp/checkpoint.h
```diff
@@ -0,0 +1,324 @@
...
+        for (auto &vec: v){
+            ReadData(parent, vec, "ind"+std::to_string(c));
+            ++c;
+        }
+    }
+};
+
+}  // namespace checkpoint_utils
+
+class CheckpointFile {
```
JensWehner: separate file

CMakeLists.txt
pmadhikar: Okay, done. Thanks!

include/votca/xtp/checkpoint.h
pmadhikar: iff as in If-and-only-if :)

CMakeLists.txt
```diff
@@ -90,6 +90,20 @@ if(CERES_FOUND)
   include_directories(${CERES_INCLUDE_DIRS})
 endif(CERES_FOUND)
 
+find_package(HDF5 1.8 REQUIRED COMPONENTS CXX)
+
+if(HDF5_FOUND)
```
junghans: With `REQUIRED` hdf5 is always found, so remove this conditional.

CMakeLists.txt
```diff
@@ -90,6 +90,20 @@ if(CERES_FOUND)
...
+  message("-- HDF5 found for checkpointing")
```
junghans: CMake already prints a message that hdf5 was found, so why do we need two?

pmadhikar: Fair enough, deleted.

CMakeLists.txt
pmadhikar: I'm just following the same way that it was done for the other packages, see libxc, ceres, eigen above.

src/tests/test_hdf5.cc
```diff
@@ -0,0 +1,189 @@
...
+    {
+        // Write orbitals
+        Orbitals orbWrite;
+
+        orbWrite.setBasisSetSize(basisSetSize);
+        orbWrite.setNumberOfLevels(occupiedLevels, unoccupiedLevels);
+        orbWrite.setNumberOfElectrons(numElectrons);
+        orbWrite.MOEnergies() = moeTest;
+        orbWrite.MOCoefficients() = mocTest;
+        orbWrite.QMAtoms() = atomsTest;
```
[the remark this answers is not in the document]
pmadhikar: Found sour of error, it was due to me not understanding how orbitals deals with the list of QMAtoms.

CMakeLists.txt
pmadhikar: Okay, I've gone ahead and removed the checks for hdf5 and the other libraries.

PR #123, opened 2018-05-02
added std:: declarations and parallelized bottleneck

JensWehner:


src/libxtp/numerical_integration/numerical_integrations.cc
```diff
@@ -748,19 +748,18 @@ void NumericalIntegration::GridSetup(string type, vector<QMAtom*> _atoms,const A
           } // if ( ait != bit) 
           i_b++;
         }// bit centers
-
+#pragma omp parallel for schedule(guided)
         for (unsigned i_grid = 0; i_grid < _atomgrid.size(); i_grid++) {
           // call some shit called grid_ssw0 in NWChem
```
baumeier: we might want to adjust the comments a bit ;)

JensWehner: I know who has this kind of prose in their code. :D

PR #128, opened 2018-05-23
Refactor esp

JensWehner:
Refactored espfit a bit and added unittest.


include/votca/xtp/espfit.h
```diff
@@ -60,12 +60,12 @@ class Espfit{
...
      
-    double getNetcharge( std::vector< QMAtom* >& _atoms, double N );
+    double getNetcharge(const std::vector< QMAtom* >& _atoms, double N );
  
-    Eigen::VectorXd EvalNuclearPotential( std::vector< QMAtom* >& _atoms, Grid _grid );
+    void EvalNuclearPotential(const std::vector< QMAtom* >& _atoms, Grid& _grid );
    
      // Fits partial charges to Potential on a grid, constrains net charge
-    std::vector<double> FitPartialCharges( std::vector< tools::vec >& _fitcenters, Grid& _grid, Eigen::VectorXd& _potential, double& _netcharge );
+    void FitPartialCharges(std::vector< QMAtom* >& _atoms,const Grid& _grid, double _netcharge );
```
baumeier: You replaced _fitcenters by _atoms, which does not *have* to be the case.

JensWehner: agreed, I reduce the functionality here, if we ever want to add virtual sites we have to reimplement that, at the moment I think though the reduced complexity is better.

PR #132, opened 2018-06-15
Rework includes

JensWehner:
- reduces namespace poisoning, removed using namespace tools from aoshell.h
- fixed all dependencies
- replaced homemade erfc function in numerical integration by std::erfc(std::abs())
- changed TC_gwbse ordering from \[mos\]\(aux,mos\) to \[mos\]\(mos,aux\) because 2nd dimension is fasted and sigma is about 20% faster this way
- fixed threecenter unittest


JensWehner:
:)

PR #136, opened 2018-06-29
changed how charges are handled #134

### LINKED ISSUE #134: QMMM writing of coordinates is broken
Independent of qmpackage, no coordinates are written into the DFT package input files.

JensWehner:
The polarsegments are now set via the setMultipoles, 

if the multipoles are set, the qmpackages automatically take care that the correct option is added.

Fix #134


baumeier:
How does this behave if std::vector<ctp::PolarSeg*> PolarSegments is empty? So for cutoff=0?

JensWehner:
at the moment, it writes zero charges, so basically the reaction depends on the DFT code. What do we want instead?

PR #144, opened 2018-08-03
should add electron when occ != 0

mbarbry:
I think I caught an error in the nwchem parser.
The parser adds electron only if the occupation number is equal to two. But I generated calculations were these numbers were equal to one. Checking with the main log file, it thus makes sense to me that electrons should be counted if the occupation number is different than 0.

But I don't know so well nwchem, it should be checked.

### DIFF
--- a/src/libxtp/qmpackages/nwchem.cc
+++ b/src/libxtp/qmpackages/nwchem.cc
@@ -535,7 +535,7 @@ namespace votca {
             for (i = 1; i <= _n_lines; i++) {
                 for (int j = 0; j < 3; j++) {
                     _input_file >> _occ[ _imo ];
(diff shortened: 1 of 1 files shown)


JensWehner:
I agree, but for this we have to extend our whole infrastructure for open shell systems and till then I ll wait

baumeier:
Just to confirm:  When you run a closed shell system, things are working as they should?

If this occurs for open-shell calculations, I agree with @JensWehner to wait until we have proper structures to read MO energies and coefficients into.

baumeier:
Basically needs to be addressed with #6

junghans:
Needs update!

JensWehner:
as it is on the feature list, this will be closed for the moment

PR #150, opened 2018-08-22
.travis.yml: only test some branches

junghans:
Once that works we should add the same block to all other VOTCA repos.


junghans:
Also we might also want to allow all branches starting with`omg-I-know-what-I-am-doing`.

Issue #156, opened 2018-09-15
Output of excited state dipole moment (or higher moments)


baumeier:
Analysis of dipole moment of QM density needs to be available somewhere. Maybe in esp2multipole as reference for the fit.

PR #160, opened 2018-09-18
Output of dipoles for states

JensWehner:
- closing issue https://github.com/votca/xtp/issues/156
- made BSE errors more explicit
- made AOMatrix::FillBlock a pure virtual function, e.g it has to be implemented by derived class
- converted some member variables to lower case
- aodipole now has a method to set the center around which the dipole can be calculated, defaults to origin
-fixed bug in bse fragment analysis, before it was always executed.


felipeZ:
@JensWehner I like the modern syntax with which you have replace the C++98 syntax.

PR #161, opened 2018-09-24
Replace ceres by internal

JensWehner:
- Refactored bfgs-trm into own class
- Function to optimize has been decoupled from the optimizer
- much tighter tolerances than before
- ability to set initial guess for H
- less chance to run into infinite loops
- more unit_tests


junghans:
Can we remove ceres from buildenv as well then?

JensWehner:
yes

JensWehner:
I would like to have a bit user experience though.

PR #162, opened 2018-09-24
Fix reading and writing bool to hdf5 by casting to/from int

### LINKED ISSUE #159: HDF5 bool support
HDF5 natively supports no bools, 

we have to convert them to int and back.

pmadhikar:
Now we simply cast before and after the read. Just as before the
callers define the type, so it works the same as before but with a
specialized functions dealing with bools.

Fix #159


pmadhikar:
Turned out to be a simple fix :)

PR #170, opened 2018-10-11
Hdf5 errorcheck

pmadhikar:
This contains changes that adds: 

- Error checking and tests

- Accessor types for the checkpoint file

- Checkpoint file backups if one already exists (no more overwrite by default)


src/libxtp/orbitals.cc
```diff
@@ -634,171 +634,157 @@ namespace votca {
...
+            w(_unoccupied_levels, "unoccupied_levels");
+            w(_number_of_electrons, "number_of_electrons");
+
+            w(_mo_energies, "mo_energies");
+            w(_mo_coefficients, "mo_coefficients");
+
+            // write qmatoms
+
+            {
+                size_t count = 0;
```
JensWehner: instead of a count please use for(unsigned=0 etc....

include/votca/xtp/checkpoint.h
```diff
@@ -20,24 +20,40 @@
...
 #include <votca/xtp/checkpoint_utils.h>
-
+#include <votca/xtp/checkpointwriter.h>
+#include <votca/xtp/checkpointreader.h>
 
 namespace votca {
 namespace xtp {
 
+enum class CheckpointAccessLevel{
+        READ = 0,
```
JensWehner: use slightly different names,READ, CREATE,SAVECREATE maybe?

PR #178, opened 2018-11-23
Optimiser fix

JensWehner:
- orbitals object can be safely copied now
- aobasis object can be safely copied now
- forces object does not modify orbitals object anymore
- pointer misplacement, when writing and then reading orb file has been fixed
- optimiser now prints initial config at iteration 0
- reordered how iterations are counted in the optimiser
- all energies in orbitals object are now in Hartree no more eV
- removed "noisy" option in forces and instead the -v in the calculator puts this out
- Run() function in qmpackage no longer needs an orbitals input argument,


baumeier:
@JensWehner please take care of the orbitals conflict

PR #183, opened 2018-12-21
Sigma refactor

JensWehner:
- Sigma can now be exchanged for each case, new sigma implementations derive from sigma_base
- RPA rewritten to be better suited for integration
- more tests
- removed  GW0 formalism
- inversion of auxmatrices is now handled by threecenter matrices
- BSE is completely decoupled from the sigma implementation
- PPM is now part of sigma_ppm


junghans:
Travis fails with:
```
[ 90%] Building CXX object xtp/src/tests/CMakeFiles/unit_test_bse.dir/test_bse.cc.o
[91m/home/votca/votca/xtp/src/tests/test_bse.cc:21:10: fatal error: votca/xtp/sigma.h: No such file or directory
 #include <votca/xtp/sigma.h>
          ^~~~~~~~~~~~~~~~~~~
compilation terminated.
```

Did you forget to add `sigma.h`?

JensWehner:
Thanks, I did not change the include properly.

PR #192, opened 2019-01-03
Clean namespace from using

JoshuaSBrown:
Any source files that were missing the using namespace std; line were addressed so that when the header files in tools are fixed it does not create problems for xtp.


JoshuaSBrown:
@JensWehner I'm not entirely sure if I should be merging this with master or this branch?

PR #193, opened 2019-01-03
Clean namespace from using2

JoshuaSBrown:


JoshuaSBrown:
I think I did this the wrong way around.

JoshuaSBrown:
Will just wait until fix-4 is merged, and these commits will come with the merge.

PR #202, opened 2019-01-15
fixed bug in dft convergence 4c integrals and added more unit tests

JensWehner:
- fixes two bugs
- one in eris, the offdiagonal elements of exx where computed twice
- the hist length for the diis was not set
- added more unit tests
- removed all the pointer stuff from adiis and diis as it is not worth it
- removed print functions from matrix objects
- slightly improved dft output

found by @NicoRenaud


share/packages/mbgft.xml
```diff
@@ -1,7 +1,7 @@
 <gwbse>
         <dftbasis>ubecppol</dftbasis>
         <gwbasis>aux-ubecppol</gwbasis>
-        <ranges></ranges> <!-- default: all levels in RPA, 1:2*HOMO in QP and BSE; other options: factor,explicit -->
+        <ranges>default</ranges> <!-- default: all levels in RPA, 1:2*HOMO in QP and BSE; other options: factor,explicit -->
```
baumeier: Are the defaults still this way? Should we change them?

JensWehner: if you want, the are still [0,homo][homo,2xhomo]

PR #205, opened 2019-01-17
Update development with master

JensWehner:


include/votca/xtp/adiis.h
```diff
@@ -24,24 +24,21 @@
...
  
  class ADIIS{
 public:
-
-    ADIIS():success(true) {};
-   ~ADIIS() {};
-   
-    Eigen::VectorXd CalcCoeff(const std::vector< Eigen::MatrixXd* >& _dmathist,const std::vector< Eigen::MatrixXd* >& _mathist);
+  
+    Eigen::VectorXd CalcCoeff(const std::vector< Eigen::MatrixXd >& dmathist,const std::vector< Eigen::MatrixXd >& mathist);
```
felipeZ: It is great that you are getting read of pointers!!

include/votca/xtp/bse.h
```diff
@@ -56,31 +42,34 @@ struct Population {
...
-                _bse_vtotal = _bse_vmax - _bse_vmin + 1;
-                _bse_ctotal = _bse_cmax - _bse_cmin + 1;
-                _bse_size = _bse_vtotal * _bse_ctotal;
-                return;
-            }
+ 
 
-   
-  void Solve_triplets();
+   void configure(const options& opt){
```
felipeZ: I think that the use `options` make the code more robust by isolating encapsulating the options in a single entity

include/votca/xtp/convergenceacc.h
```diff
@@ -25,112 +25,102 @@
...
-    
+
     enum KSmode { closed, open, fractional };
 
-    ConvergenceAcc() {_mode=KSmode::closed;
-                       _usemixing=true;
-                      _diiserror=std::numeric_limits<double>::max();
-                      _maxerrorindex=0;
-                      _maxerror=0.0;};
+    struct options{
```
felipeZ: This make the code a lot simpler and easier to use, great!

include/votca/xtp/threecenter.h
```diff
@@ -115,27 +110,28 @@ namespace votca {
...
-            void Print(std::string ident);
             void Fill(const AOBasis& auxbasis, const AOBasis& dftbasis, const Eigen::MatrixXd& dft_orbitals);
+            //Rebuilds ThreeCenterIntegrals, only works if the original basisobjects still exist
+            void Rebuild(){
+                Fill(*_auxbasis,*_dftbasis,*_dft_orbitals);
+            }
 
             void MultiplyRightWithAuxMatrix(const Eigen::MatrixXd& AuxMatrix);
 
-            void Cleanup();
```
felipeZ: A naive question, why is this cleanup method remove?

JensWehner:
I disagree. 

In general no` auto`. Why? 
a) Eigen3 and `auto `do not really work well together http://eigen.tuxfamily.org/dox/TopicPitfalls.html 
b) It often makes code harder to read, because I have to look up what auto actually is. 
c) for template programming it is sometimes a must
d) for obvious things or long types I do not mind it, but sometimes a good typedef is much better
e) for return values of functions it should be forbidden

This posts sums it up pretty well
https://stackoverflow.com/questions/6434971/how-much-is-too-much-with-c11-auto-keyword

So I, in general think as little as possible, as much as necessary. 

f) I want to encourage people to think about their datatypes and `auto` does not do that. I find often when people use `auto`, they also do not care about `const` or `&` either.

JensWehner:
Why do you think it should be encouraged? I mean I am more often wrong than right

include/votca/xtp/threecenter.h
JensWehner: no not naive, a lot of the classes have still methods like free matrix or cleanup. Thiis normally has one of two reasons a) the class has a reasonably big matrix in it which takes up space b) Something with pointers. Now the second case should not be relevant here anymore. For the first case we have a lot shorter functions and classes now, so objects fall out of scope a lot quicker.  Secondly, our systems are larger now and only very few objects actually matter in size due to the scaling. Thirdly, I prefer I bit more memory and shorter code to situations where cleanup has already been called and someone accesses an invalid object

PR #209, opened 2019-01-23
turned sm1 into S and fixed the fragment problem for the nuclear charges

JensWehner:
Fixed the bug with fragment analysis.


src/libxtp/orbitals.cc
```diff
@@ -479,7 +479,7 @@ Eigen::MatrixXd Orbitals::CalcAuxMat_cc(const Eigen::VectorXd& coeffs)const{
             Eigen::VectorXd fragmentCharges = Eigen::VectorXd::Zero(2);
             Eigen::SelfAdjointEigenSolver<Eigen::MatrixXd> es;
             es.compute(overlapmatrix);
-            Eigen::MatrixXd sqrtm1 = es.operatorInverseSqrt();
+            Eigen::MatrixXd sqrtm1 = es.operatorSqrt();
```
baumeier: If it is no longer the inverse sqrt of S, should the variable name not also reflect that change?

JensWehner: good point

PR #211, opened 2019-01-24
Ignore corelevels

JensWehner:
make the idea of ignoring corelevels more consistent


src/libxtp/gwbse/gwbse.cc
```diff
@@ -124,6 +124,7 @@ void GWBSE::Initialize(tools::Property& options) {
     int ignored_corelevels = 0;
     if (ignore_corelevels) {
         ignored_corelevels = CountCoreLevels();
+        rpamin=ignored_corelevels;
```
baumeier: Does this make a difference in the results?

PR #214, opened 2019-01-27
Add magic for stable testing to gitlab-ci

junghans:


.gitlab-ci.yml
```diff
@@ -0,0 +1,111 @@
...
+      - ccache/
+  script:
+    - ccache -z
+    - j="$(grep -c processor /proc/cpuinfo 2>/dev/null)" || j=0; ((j++))
+    - git checkout -B current_commit
+    - cd $HOME
+    - git clone --depth=1 --recursive https://github.com/votca/votca
+    - pushd votca
+    - if [[ ${CI_COMMIT_REF_NAME} = stable* ]]; then
+        git checkout stable && git submodule update --recursive --init;
```
junghans: This is the crux.

junghans:
This will only work for `stable`, and not e.g. `fix-4`, so maybe it would be better to do something more general, e.g. branchname `for_<XXX>_<more_info> will trigger a build on top the `XXX` branch (if that branch exists).

junghans:
We could also use `for/<branch>/topic` as a branch name, slashes are more descriptive!

junghans:
Merge #215 instead .

PR #215, opened 2019-01-27
.gitlab-ci.yml: add branch magic

junghans:
Fix #214


junghans:
The new logic is that if you name your branch `for/stable/something` it will build it against stable instead of master! For master all branch names works! The new logic also applies to other branches, e.g. `for/fix-4/something` works as well!

PR #219, opened 2019-01-29
orca parses correct final single point energy in case of dispersion c…

baumeier:
…orrection

### DIFF
--- a/src/libxtp/qmpackages/orca.cc
+++ b/src/libxtp/qmpackages/orca.cc
@@ -482,10 +482,10 @@ namespace votca {
                     }
                 }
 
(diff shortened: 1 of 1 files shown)


JensWehner:
What happens if you have outside charges?

baumeier:
Is not added. Orca only adds the electrostatic environment energy, if the charges are defined in the XYZ block, as Q, q, x,y,z. If given in a separate pointcharge file, as we do, it reports the pure electronic energy:

"Note that when adding point charges by reading them through a pointcharge file, the self-interaction between point charges is NOT calculated (unlike above)."

See https://sites.google.com/site/orcainputlibrary/geometry-input

PR #232, opened 2019-02-07
Pcc64le fix lexical cast

JoshuaSBrown:


junghans:
Please refrain from running clang-format and friends on the `stable` code base, this is just leads to endless merge conflict, which @JensWehner now was to solve, when merging with master or fix-4.

PR #235, opened 2019-02-09
cmake: bump version to 1.6-dev

junghans:


.gitlab-ci.yml
```diff
@@ -11,19 +11,26 @@ stages:
...
     - ccache -z
     - j="$(grep -c processor /proc/cpuinfo 2>/dev/null)" || j=0; ((j++))
     - git branch commit_of_build_${CI_BUILD_ID} ${CI_COMMIT_SHA}
-    - cd $HOME
-    - git clone --recursive https://github.com/votca/votca
+    - git clone https://github.com/votca/votca
     - pushd votca
-    - if [[ ${CI_COMMIT_REF_NAME} =~ for/([^/]*)/.* && ${BASH_REMATCH[1]} != master ]]; then
-        git checkout -b ${BASH_REMATCH[1]} origin/${BASH_REMATCH[1]} && git submodule update --recursive --init;
+    - if [[ ${CI_COMMIT_TAG} = v[12].[0-9]* ]]; then
```
JoshuaSBrown: What exactly is this code block doing? I understand the else statement, but what is the if statement doing. Are we supporting merging with different versions ?

junghans: If we are build for a tag, we use a slightly different checkout command!

junghans: And a tag is triggered by a release!

junghans: This code is already on the stable branch!

PR #237, opened 2019-02-14
Electrostatics

JensWehner:
Better integration of electrostatics with DFT

Part of votca/votca#175 and https://github.com/votca/votca/pull/178


JoshuaSBrown:
Which files did you primarily change, I see that you are using the formatter which makes it a little harder to determine which part of the code you changed.

PR #242, opened 2019-02-25
Bse operator jens

JensWehner:
- Removed float option, everything is double now
- fixed the options problem
- fixed a couple of small errors
- made a couple of variables private in matrix_freeoperator
- merged with master


JensWehner:
Argh the branch before was not formatted shit.

PR #254, opened 2019-03-25
Increase H_x speed. This is not tested yet

JensWehner:
@NicoRenaud maybe this helps speed up the bse a bit. It should allocate a lot less memory.


NicoRenaud:
Tested on the benchmark set. Gives the same eigenvalues with a 15% speed improvement

Issue #229, opened 2019-02-06
test_espfit takes to long to run


JoshuaSBrown:
Move unit_test_espfit to a different flag from the unit tests, maybe to regression test...

JensWehner:
Is this urgent, otherwise I would fix it later. The reason is a comparison between numeric and analytic integration

PR #256, opened 2019-03-26
shortened unit test

JensWehner:
reduced size of the espfit unit test should fix https://github.com/votca/xtp/issues/229


JensWehner:
@JoshuaSBrown I need a her.... review.

PR #260, opened 2019-04-10
Btdatriplets

JensWehner:
- Implements BTDA for Triplets
- Caching Hx results to speed up calculation
- bsecoupling uses BSE operators now for projection
- Restart of BSE calculation from GW results possible
- Diagonalisation of Hqp is always done and saved in orb file
- Orbitals file cannot save BSE Hamiltonian anymore


include/votca/xtp/bse.h
```diff
@@ -146,13 +146,24 @@ class BSE {
...
+
   template <typename BSE_OPERATOR>
   void configureBSEOperator(BSE_OPERATOR& H);
 
   template <typename BSE_OPERATOR>
   void solve_hermitian(BSE_OPERATOR& H, Eigen::VectorXd& eigenvalues,
                        Eigen::MatrixXd& coefficients);
 
+  template <typename BSE_OPERATOR_ApB, typename BSE_OPERATOR_AmB>
+  void Solve_antihermitian(BSE_OPERATOR_ApB& apb, BSE_OPERATOR_AmB& amb,
```
NicoRenaud: just to be annoying : `solve_` instead of `Solve_` :)

include/votca/xtp/orbitals.h
```diff
@@ -307,6 +288,14 @@ class Orbitals {
     return _BSE_triplet_coefficients;
   }
 
+  const Eigen::MatrixXd& BSETripletCoefficientsAR() const {
+    return _BSE_triplet_coefficients_AR;
+  }
+
+  Eigen::MatrixXd& BSETripletCoefficientsAR() {
```
NicoRenaud: I guess the answer is yes, but do you need both ?

src/libxtp/gwbse/bse_operator.cc
```diff
@@ -46,21 +46,36 @@ Eigen::RowVectorXd BSE_OPERATOR<cqp, cx, cd, cd2>::row(int index) const {
...
   int auxsize = _Mmn.auxsize();
   vc2index vc = vc2index(0, 0, _bse_ctotal);
-  Eigen::VectorXd Hrow = Eigen::RowVectorXd::Zero(_bse_size);
+
   const int vmin = _opt.vmin - _opt.rpamin;
   const int cmin = _bse_cmin - _opt.rpamin;
   int v1 = vc.v(index);
   int c1 = vc.c(index);
-  const Eigen::VectorXd Mmn1Tc1 = _Mmn[v1 + vmin].row(c1 + cmin);
+  int cache_size = 10;
```
NicoRenaud: could we make the cache size an option somehow ? Let say we have huge matrices and we don't want to cache, would be nice to set the size to 0.

NicoRenaud:
Looks great !

include/votca/xtp/bse.h
JensWehner: true, I merged before I saw the comments, I ll fix it in the main version.

include/votca/xtp/orbitals.h
JensWehner: yes, one is the const version the other is not. So yeah

src/libxtp/gwbse/bse_operator.cc
JensWehner: Yeah that is a good point. So I briefly benchmarked a bit and after 10 the effect seemed quite small. The memory need for 10 eigenvectors is actually also not that large. So I would argue, that the option confuses more than is actually helpful. I understand your point, but I would leave it as it is, if the need at some point arises, adding it is no problem

PR #266, opened 2019-05-17
Electrostatics joshs development [WIP]

JoshuaSBrown:
Introduced Unit Conversion class from tools to handle conversion between csg and xtp topology classes
Introduced generic structure parameters class from tools to create generic constructors for helping with more generalized template methods 
Introduced file adapter to make use of all of the file readers and writers in csg by converting xtp objeccts first to csg topology objects.


include/votca/xtp/fileadapter.h
```diff
@@ -0,0 +1,211 @@
...
+     * This class is designed to interface with the topology structures in xtp
+     * and in csg such that the xtp classes can use the csg topolology and 
+     * trajectory input and output readers and writers
+     */
+    class FileAdapter {
+
+      public: 
+
+        template<class XTP_Topology_T, class CSG_Topology_T>
+          void convertXTPTopologyToCSGTopology(XTP_Topology_T & from_xtp_top, CSG_Topology_T & to_csg_top);
```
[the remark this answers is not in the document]
JensWehner: i.e. xtp::topology convertToXTP(const CSG_Topology& top)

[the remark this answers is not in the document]
JensWehner: should the methods itself be const?

include/votca/xtp/fileadapter.h
```diff
@@ -0,0 +1,211 @@
...
+
+
+    /**
+     * @brief File Topology Adapter
+     *
+     * This class is designed to interface with the topology structures in xtp
+     * and in csg such that the xtp classes can use the csg topolology and 
+     * trajectory input and output readers and writers
+     */
+    class FileAdapter {
```
JensWehner: This is a topology converter not a FileAdapter

include/votca/xtp/fileadapter.h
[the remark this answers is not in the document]
JensWehner: method can be called Convert if you use the proper return types etc...

include/votca/xtp/fileadapter.h
```diff
@@ -0,0 +1,211 @@
...
+          /// Set the molecule id and type equal to the atom container id and type
+          typename CSG_Topology_T::container_t  mol = to_csg_top.CreateMolecule(params);  
+          for( typename XTP_Topology_T::bead_t & bead : xtp_cont ){
+
+            tools::StructureParameters params_bead = bead.getParameters();
+            Eigen::VectorXd pos = params_bead.get<Eigen::VectorXd>(tools::StructureParameter::XTP_Position);
+            pos*= converter_.convert(XTP_Topology_T::distance_unit,CSG_Topology_T::distance_unit);
+            params_bead.set(tools::StructureParameter::CSG_Position,pos);
+            typename CSG_Topology_T::bead_t csg_bead = to_csg_top.CreateBead(params);
+            mol.AddBead(&csg_bead);
```
JensWehner: why add a pointer?

include/votca/xtp/fileadapter.h
```diff
@@ -0,0 +1,211 @@
...
+          params.set(tools::StructureParameter::CSG_Position,pos);
+          typename CSG_Topology_T::bead_t csg_bead = to_csg_top.CreateBead(params);
+          mol.AddBead(&csg_bead);
+        }
+      }
+
+    /*****************************************************************************
+     * Converting from CSG to XTP 
+     *****************************************************************************/
+    template<class CSG_Topology_T, class Container_T>
```
JensWehner: why a template for csg_topology?

include/votca/xtp/fileadapter.h
```diff
@@ -0,0 +1,211 @@
...
+
+        tools::UnitConverter converter_;
+
+    };
+
+    /*****************************************************************************
+     * Converting from XTP to CSG 
+     *****************************************************************************/
+
+    template<class XTP_Topology_T, class CSG_Topology_T>
```
JensWehner: why all the templates?

include/votca/xtp/fileadapter.h
```diff
@@ -0,0 +1,211 @@
...
+        typename CSG_Topology_T::container_t csg_cont = from_csg_top.getMolecule(csg_molecule_id);
+        tools::StructureParameters params = csg_cont.getParameters();
+        /// Set the molecule id and type equal to the atom container id and type
+        int mol_id = params.get<int>(tools::StructureParameter::MoleculeId);
+        std::string molecule_type = params.get<std::string>(tools::StructureParameter::MoleculeType);
+        std::unordered_map<int,std::vector<tools::StructureParameters>> mol_id_and_bead_params;
+
+        std::vector<int> bead_ids = csg_cont.getBeadIds();
+        for( const int & bead_id : bead_ids ){
+          typename CSG_Topology_T::bead_t * bead = csg_cont.getBead(bead_id);
```
JensWehner: Do not return a pointer but a reference

include/votca/xtp/fileadapter.h
```diff
@@ -0,0 +1,211 @@
...
+          Eigen::VectorXd pos = bead_params.get<Eigen::VectorXd>(tools::StructureParameter::CSG_Position);
+          pos*= converter_.convert(CSG_Topology_T::distance_unit,Container_T::distance_unit);
+          bead_params.set(tools::StructureParameter::XTP_Position,pos);
+          typename Container_T::bead_t bead(bead_params);
+          to_cont.push_back(bead);
+        }
+
+      }
+
+    template<class CSG_Topology_T, class XTP_Topology_T>
```
JensWehner: remove templates

include/votca/xtp/fileadapter.h
```diff
@@ -0,0 +1,211 @@
...
+        to_xtp_top.setBox(box);
+
+        to_xtp_top.setStep(from_csg_top.getStep());
+        to_xtp_top.setTime(from_csg_top.getTime() * 
+            converter_.convert(CSG_Topology_T::time_unit,XTP_Topology_T::time_unit));
+
+        std::vector<std::pair<int,std::string>> id_and_types;
+        std::unordered_map<int,std::vector<tools::StructureParameters>> mol_id_and_bead_params;
+        /// We must collect the molecules and beads in order because the 
+        /// order that they are created in xtp is important.
```
JensWehner: why? I do not understand the comment

include/votca/xtp/md2qmengine.h
```diff
@@ -18,7 +18,7 @@
 #ifndef VOTCA_XTP_MD2QMENGINE_H
 #define VOTCA_XTP_MD2QMENGINE_H
 
-#include <votca/csg/topology.h>
+#include <votca/csg/csgtopology.h>
```
JensWehner: why the renaming they are in separate namespaces

include/votca/xtp/md2qmengine.h
```diff
@@ -28,15 +28,15 @@ class Md2QmEngine {
...
+  Topology map(const csg::CSG_Topology& top);
 
  private:
   void CheckMappingFile(tools::Property& topology_map) const;
   template <class T>
   bool SameValueForMultipleEntries(const std::vector<tools::Property*>& props,
                                    std::string tag) const;
 
-  int DetermineResNumOffset(const csg::Molecule* mol,
+  int DetermineResNumOffset(const csg::Molecule& mol,
```
JensWehner: good catch

include/votca/xtp/mmregion.h
```diff
@@ -44,7 +45,7 @@ class MMRegion : public Region {
     return _segments.end();
   }
 
-  void WritePDB(csg::PDBWriter& writer) const;
+  void WritePDB(csg::PDBWriter<csg::CSG_Topology>& writer) const;
```
JensWehner: why the template?

include/votca/xtp/qmmolecule.h
```diff
@@ -22,14 +22,25 @@
...
+#include <votca/tools/unitconverter.h>
 
 namespace votca {
 namespace xtp {
 
 class QMMolecule : public AtomContainer<QMAtom> {
  public:
+  typedef QMAtom bead_t;
+
+  static const tools::DistanceUnit distance_unit;
```
JensWehner: what is this?

include/votca/xtp/topology.h
```diff
@@ -45,6 +62,9 @@ class Topology {
   // I do not have to manually make a move constructor or move assignment
   // operator or destructor because I only have to reassign pointers in qmnblist
   // object
+//  size_t BeadCount() const;
```
JensWehner: remove

include/votca/xtp/topology.h
```diff
@@ -36,6 +38,21 @@ class Segment;
  */
 class Topology {
  public:
+
+  typedef Segment::bead_t bead_t;
```
JensWehner: what is this?

src/libxtp/atom.cc
```diff
@@ -17,10 +17,18 @@
  *
  */
 #include "votca/xtp/atom.h"
+#include <votca/tools/types.h>
 
 namespace votca {
 namespace xtp {
 
+  Atom::Atom(tools::StructureParameters params) :
```
JensWehner: I do not get this.

src/libxtp/md2qmengine.cc
```diff
@@ -143,27 +143,27 @@ Topology Md2QmEngine::map(const csg::Topology& top) {
...
     // we have to figure out how the Residue numbers change from molecule to
     // molecule to get the correct mapping information this does not require for
     // the atoms to be sorted according to resnum
     int ResNumOffset = DetermineResNumOffset(mol, resnums_map);
-    for (const csg::Bead* bead : mol->Beads()) {
+    for (const csg::Bead* bead : mol.getBeads()) {
       Segment* seg =
-          segments[MolToSegMap[mol->getName()][bead->getResnr() - ResNumOffset]
-                              [bead->getName()]];
+          segments[MolToSegMap[mol.getType()][bead->getResidueId() - ResNumOffset]
```
JensWehner: what is the difference between type and name?

src/libxtp/mmregion.cc
```diff
@@ -37,7 +37,7 @@ std::string StaticRegion::identify() const {
 }
 
 template <class T>
-void MMRegion<T>::WritePDB(csg::PDBWriter& writer) const {
+void MMRegion<T>::WritePDB(csg::PDBWriter<csg::CSG_Topology>& writer) const {
```
JensWehner: is there any other writer than CSG_Topology?

src/libxtp/qmmolecule.cc
```diff
@@ -16,33 +16,58 @@
...
 
 using namespace std;
 using namespace votca::tools;
 
 namespace votca {
 namespace xtp {
 
+const tools::DistanceUnit QMMolecule::distance_unit =                             
+    tools::DistanceUnit::angstroms;                                             
+const tools::MassUnit QMMolecule::mass_unit =
```
JensWehner: why do you have to define it for every class and not simply generic for votca xtp?

src/libxtp/qmmolecule.cc
```diff
@@ -16,33 +16,58 @@
...
+    return params;
+  }
+
 void QMMolecule::WriteXYZ(std::string filename, std::string header) const {
-  csg::XYZWriter writer;
+  csg::XYZWriter<csg::CSG_Topology> writer;
   writer.Open(filename, false);
-  writer.Write(*this, header);
+  writer.WriteHeader(header,size());
+  FileAdapter file_adapter;
```
JensWehner: why is FileAdapter even a class it only has static functions?

src/libxtp/qmmolecule.cc
```diff
@@ -16,33 +16,58 @@
...
 
 void QMMolecule::LoadFromFile(std::string filename) {
-  csg::XYZReader reader;
-  reader.Open(filename);
-  reader.ReadFile<QMMolecule>(*this);
-  reader.Close();
+  csg::XYZReader<csg::CSG_Topology> reader;
+  //reader.Open(filename);
+  csg::CSG_Topology csg_temp;
+  reader.ReadTopology(filename,csg_temp);
```
JensWehner: I do not like the temp stuff

JensWehner: I had it before but better is *this=bla.readfile(filename=

src/libxtp/qmmolecule.cc
JoshuaSBrown: This was a design decision I had two options, if I define it in every class then whenever I pass in the class to the unit converter by default it knows how to convert it, I don't have to worry about passing in extra options to specify the units and possibly messing it up. Plus if the units are declared static const there should only be one instantiation per class so it should not increase the memory (I don't know if that was done for the above case will have to look). 

What do you think? 

Maybe declare all the default units used in xtp and package them in a single struct and just include the struct as a static const member variable of the classes/ or equivalent. I don't think it should be a performance problem if done correctly because it should be possible to evaluate the units at compile time.

src/libxtp/qmmolecule.cc
JoshuaSBrown: Ok Will see about changing.

src/libxtp/mmregion.cc
JoshuaSBrown: Not right now, but I went through the trouble of making it flexible prior to the changes you made in xtp. At this point I could leave it or revert it, Having already done all the work I'm a little hesitant to take the time to revert it unless we are struggling to keep our compile times down.

include/votca/xtp/fileadapter.h
JoshuaSBrown: True enough will fix names.

include/votca/xtp/fileadapter.h
JoshuaSBrown: Good point will fix.

src/libxtp/qmmolecule.cc
JoshuaSBrown: What would you suggest instead standalone functions?

src/libxtp/qmmolecule.cc
JoshuaSBrown: Ok, this would prove a real headache to change, because currently these methods return a bool which is used in several parts of csg and elsewhere. At the moment the cost of changing this is high and the reward is low. We can create an issue if it really bugs you but it's probably not worth pursuing.

include/votca/xtp/fileadapter.h
JoshuaSBrown: Need to fix.

include/votca/xtp/fileadapter.h
JoshuaSBrown: Good point.

include/votca/xtp/fileadapter.h
JoshuaSBrown: ok

include/votca/xtp/fileadapter.h
JoshuaSBrown: ok

include/votca/xtp/fileadapter.h
JoshuaSBrown: xtp assigns ids to the atoms in the order they are created, so for the ids of the beads in csg to line up with the ids of atoms in xtp they first needed to be ordered.

include/votca/xtp/md2qmengine.h
JoshuaSBrown: Will revert.

include/votca/xtp/fileadapter.h
```diff
@@ -0,0 +1,211 @@
...
+        for( typename XTP_Topology_T::container_t & xtp_cont : from_xtp_top){
+          tools::StructureParameters params = xtp_cont.getParameters();
+          params.set(tools::StructureParameter::MoleculeId,params.get<int>(tools::StructureParameter::AtomContainerId));
+          params.set(tools::StructureParameter::MoleculeType,params.get<std::string>(tools::StructureParameter::AtomContainerType));
+          /// Set the molecule id and type equal to the atom container id and type
+          typename CSG_Topology_T::container_t  mol = to_csg_top.CreateMolecule(params);  
+          for( typename XTP_Topology_T::bead_t & bead : xtp_cont ){
+
+            tools::StructureParameters params_bead = bead.getParameters();
+            Eigen::VectorXd pos = params_bead.get<Eigen::VectorXd>(tools::StructureParameter::XTP_Position);
```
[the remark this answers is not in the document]
JoshuaSBrown: ok will fix.

include/votca/xtp/mmregion.h
JoshuaSBrown: We have talked about this before, it was related to the use of the topology object in xtp before you made changes to this object. I had made the file readers and writers generic to handle topology objects from csg, xtp and even ctp if needed. At this point I don't really care to revert the changes. We can talk about it more if it bothers you.

include/votca/xtp/qmmolecule.h
JoshuaSBrown: This is used by the file readers and writers to resolve the types stores in each of the containers.

include/votca/xtp/fileadapter.h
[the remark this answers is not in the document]
JoshuaSBrown: Done

src/libxtp/md2qmengine.cc
JoshuaSBrown: Originally, the csg bead had three attributes associated with the name member, they were stored in a single string name_ with the following format:

"beadtype::redue_type::resiude_number"

This I thought was redundant and confusing, because the beads also had separate members for the residue number, so I split the pieces up into their own parts. I.e. the residue number is stored as an integer only and not in a name member, the residue type as a string and the bead type as a separate string. To get the combination one can call getLabel  this will append the different components together as a string. To be consistent with this approach the molecule method getName was also changed to getType. I don't know if I'm being clear or not...

include/votca/xtp/topology.h
JoshuaSBrown: done

src/libxtp/atom.cc
JoshuaSBrown: Can you be a little more specific? By using a standardized constructor for different objects, one can more effectively use templates. You don't have to specialize the construction of a template method for each new class you support.

include/votca/xtp/topology.h
JoshuaSBrown: A standardized label for referring to to the type stored in the container, for use with templates.

JoshuaSBrown:
Still need unit test for topology converter

PR #272, opened 2019-06-21
Trying to build against votca without building ctp

JoshuaSBrown:
Removing support for ctp through building on gitlab runner.


JoshuaSBrown:
@junghans if you have any advice on how to do this I would be indepted, this is driving me crazy. I don't understand why the lines that I changed are now causing problems with gitlabs ability to find the votca tools.

JoshuaSBrown:
Do I need the CMakeModule for VOTCA TOOLS because I tried that...?

CMakeLists.txt
```diff
@@ -82,7 +82,7 @@ if(NOT TARGET manpages)
   add_custom_target(manpages ALL)
 endif()
 
-find_package(VOTCA_TOOLS REQUIRED NO_MODULE)
+find_package(VOTCA_TOOLS NO_MODULE REQUIRED)
```
JensWehner: why that?

JensWehner:
I really have not too much expertise in this.

CMakeLists.txt
JoshuaSBrown: Umh... how about become I don't know what I'm doing with cmake and the documentation and examples are lacking even though its been around for years.

JoshuaSBrown:
The problem is that I was trying to build against votca/electrostatics which does not exist so I think it was defaulting to master. I can get around this byusing for/development/

Issue #287, opened 2019-10-01
test_aotransform: fails to compile with clang-9


junghans:
From [gitlab](https://storage.googleapis.com/gitlab-gprd-artifacts/45/a9/45a98d28d7e7b0bfcab721c1afad3cab94f5e4a87a90568c5ea0437a447357f4/2019_10_01/308305329/318985874/job.log?response-content-type=text%2Fplain%3B%20charset%3Dutf-8&response-content-disposition=inline&GoogleAccessId=gitlab-object-storage-prd@gitlab-production.iam.gserviceaccount.com&Signature=YRJ1sUnE5as7n2OglifA%2FZR4enXYKlAjHnvQMaf6MV3UiX%2FUL4dUCvQWRjhk%0AeDe6sLujMrhYRBTJIUmduOrQ5Ns3x7bzMpTY%2FX%2F2EsI06jrivHriNCuuAoOX%0A75%2B3bQ2tQQ0aREC8mGtu3iCNLQNYsS3edPMjUrK%2BUP%2BdQBO5WycFRFbwmlcA%0Abyeaj7xibP%2BciHDzCpJs1CJeztrs5%2FuEjgHWJpBhhlsqWj%2BK1d98jPrM97fb%0AeYDlWkuzwjlOjEkiwqSOnDBLLuiK7kgdy1274wQaKOC0TY%2BvMT13cmri23H3%0A6ZXxlsKxS1C5n0nZKXRUt5woa6BOOJV7z33AUKvbmA%3D%3D&Expires=1569895690):
```
Stack dump:
[...]

PR #290, opened 2019-10-04
Clang transform fix

JensWehner:
hopefully fixes clang break https://github.com/votca/xtp/issues/287


JensWehner:
changed, let us wait for the CI

PR #300, opened 2019-10-12
cmake: remove mkl link, now done in tools

junghans:


JensWehner:
MKL is a superset of BLAS, so I want this as a special case.

JensWehner:
I am an idiot, I was confused. Nevermind

junghans:
This has nothing to do with blas, we can remove this simply due to the fact that the linkage is done in tools.

PR #301, opened 2019-10-13
introduced block multiplication

JensWehner:
Before the matrixoperator could only do linewise matrix multiplication.
`for(i=0;i<rows()){`
 `matrix_out.row(i)=op.row(i)*matrix_in`
`}`

Now the operator can additionally use blockwise multiplication (with quadratic blocks):
(I left out the indeces)
`for(i=0,i<blocks_row){`
`for(j=0,j<blocks_col){`
`matrix_out.block(i)+=op.block(i,j)*matrix_in.block(j)`
`}`

This simplifies the Hx contribution by removing the need of caching. It also uses the symmetry of  the Hx. Benchmark on 56 threads on a 72771x72771 matrix showed a reduction from 2163s to 1468s.

I also renamed the `row()` operator to `OperatorRow()`, because `row()` is very close to the memberfunctions of `Eigen::MatrixXd`, which is derived from.


src/libxtp/matrixfreeoperator.cc
```diff
@@ -24,21 +24,59 @@ namespace xtp {
...
+    if (size() % blocksize != 0) {
+      throw std::runtime_error("blocksize is not a multiple of matrix size");
+    }
+    int blocks = size() / blocksize;
+
+// this is inefficient if blocks<num_ofthreads
+#pragma omp parallel for schedule(guided)
+    for (int i = 0; i < blocks; i++) {
+      Eigen::MatrixXd block = OperatorBlock(i, i);
+      D.segment(i * blocksize, blocksize) += block.diagonal();
```
NicoRenaud: why += and not =

src/tests/test_davidson.cc
```diff
@@ -68,7 +68,7 @@ BOOST_AUTO_TEST_CASE(davidson_full_matrix_fail) {
 class TestOperator : public MatrixFreeOperator {
```
NicoRenaud: We should add a test with OperatorBlock as well no ?

src/libxtp/gwbse/bse_operator.cc
```diff
@@ -77,33 +51,21 @@ Eigen::RowVectorXd BSE_OPERATOR<cqp, cx, cd, cd2>::row(int index) const {
...
 
 template <int cqp, int cx, int cd, int cd2>
-Eigen::RowVectorXd BSE_OPERATOR<cqp, cx, cd, cd2>::Hx_row(int index) const {
-  int thread_id = OPENMP::getThreadId();
-  if (_Hx_cache[thread_id].hasValue(index)) {
-    return _Hx_cache[thread_id].getValue(index);
-  }
-  int auxsize = _Mmn.auxsize();
-  vc2index vc = vc2index(0, 0, _bse_ctotal);
+Eigen::MatrixXd BSE_OPERATOR<cqp, cx, cd, cd2>::OperatorBlock(int row,
```
NicoRenaud: I'm a bit confused here. Does this means that we can use the Block only for an operator that contains only Hx ?

src/libxtp/gwbse/bse_operator.cc
```diff
@@ -32,38 +32,12 @@ void BSE_OPERATOR<cqp, cx, cd, cd2>::configure(BSEOperator_Options opt) {
...
-    return false;
-  }
-}
-
-template <int cqp, int cx, int cd, int cd2>
-Eigen::RowVectorXd BSE_OPERATOR<cqp, cx, cd, cd2>::row(int index) const {
+Eigen::RowVectorXd BSE_OPERATOR<cqp, cx, cd, cd2>::OperatorRow(
+    int index) const {
   Eigen::RowVectorXd row = Eigen::RowVectorXd::Zero(_bse_size);
-  if (cx != 0) {
```
NicoRenaud: So there is no Hx in OperatorRow at all ?! I'm a bit confused here

src/libxtp/gwbse/bse_operator.cc
NicoRenaud: Ok I guess that you can have both, and then you compute the 'rows' product first and then you add up the block product. Right ?

src/libxtp/matrixfreeoperator.cc
NicoRenaud: I guess because you first extract the diag of the row operator and then add up the diag of the block operator .

NicoRenaud:
I really like that the cache is gone, that was very obscure to me. I think I understand the logic behind the rows and blocks that are used at the same time. (I though it was either rows or blocks) We could add some test for the block in test_davidson if that increase coverage but I'm not sure it would since it's part of the bse tests

src/libxtp/gwbse/bse_operator.cc
JensWehner: yes

src/tests/test_davidson.cc
JensWehner: Yes we should add a blockoperator test

PR #299, opened 2019-10-11
Refactor davidson

NicoRenaud:
The entire davidson solver has been refactored and extended to include BTDA case

PR #302, opened 2019-10-16
added block_matrix test and moved from rand to a more deterministic m…

JensWehner:
…atrix

The QR vs GS think should be tested somewhere. Fixed in https://github.com/votca/xtp/pull/299

But we should write tests so that we test GS and QR, so we should probably also test with a 200x200 matrix


src/tests/test_davidson.cc
```diff
@@ -41,7 +43,31 @@ BOOST_AUTO_TEST_CASE(davidson_full_matrix) {
   int neigen = 10;
   double eps = 0.01;
   Eigen::MatrixXd A = init_matrix(size, eps);
+  Logger log;
+  DavidsonSolver DS(log);
+  DS.set_ortho("QR");
```
NicoRenaud: do we need QR here ?

NicoRenaud: I guess yes since that's a a small matrix

src/tests/test_davidson.cc
```diff
@@ -131,6 +152,139 @@ BOOST_AUTO_TEST_CASE(davidson_matrix_free) {
...
+  BOOST_CHECK_EQUAL(check_eigenvalues, 1);
+  if (!check_eigenvalues) {
+    std::cout << "ref" << std::endl;
+    std::cout << es.eigenvalues().head(neigen).transpose() << std::endl;
+    std::cout << "result" << std::endl;
+    std::cout << DS.eigenvalues().transpose() << std::endl;
+  }
+}
+
+class BlockOperator : public MatrixFreeOperator {
```
NicoRenaud: I also have a class called BlockOperator for the A/B block of the BTDA case. That might conflict with that class ... so maybe we have to rename one or the other

src/tests/test_davidson.cc
```diff
@@ -164,9 +318,9 @@ Eigen::MatrixXd extract_eigenvectors(const Eigen::MatrixXd &V,
   return W;
 }
 
-class BlockOperator : public MatrixFreeOperator {
+class HermitianBlockOperator : public MatrixFreeOperator {
```
NicoRenaud: ha ok .... should have read the entire PR before making the last comment :)

src/tests/test_davidson.cc
```diff
@@ -178,10 +332,12 @@ class BlockOperator : public MatrixFreeOperator {
   int _diag;
```
NicoRenaud: do we use that ?

src/tests/test_davidson.cc
JensWehner: I renamed your test class as HermitianBlock

src/tests/test_davidson.cc
JensWehner: I should read all the comments before commenting on your comments

src/tests/test_davidson.cc
JensWehner: Ahh that was the thing it always was complaining about.

PR #303, opened 2019-10-16
corrected wrong inheritance in matrixfreeoperator

JensWehner:
matrixfreeoperator inherited from the wrong type, I have no idea why it worked.

### DIFF
--- a/include/votca/xtp/matrixfreeoperator.h
+++ b/include/votca/xtp/matrixfreeoperator.h
@@ -40,9 +40,6 @@ namespace xtp {
 
 class MatrixFreeOperator : public Eigen::EigenBase<MatrixFreeOperator> {
  public:
-  typedef double Scalar;
-  typedef double RealScalar;
-  typedef int StorageIndex;
   enum {
     ColsAtCompileTime = Eigen::Dynamic,
     MaxColsAtCompileTime = Eigen::Dynamic,
--- a/include/votca/xtp/matrixfreeoperator.h
+++ b/include/votca/xtp/matrixfreeoperator.h
@@ -23,8 +23,26 @@
 namespace votca {
 namespace xtp {
 
-class MatrixFreeOperator : public Eigen::EigenBase<Eigen::MatrixXd> {
+class MatrixFreeOperator;
+}
+}  // namespace votca
+namespace Eigen {
+namespace internal {
+// MatrixReplacement looks-like a SparseMatrix, so let's inherits its traits:
+template <>
+struct traits<votca::xtp::MatrixFreeOperator>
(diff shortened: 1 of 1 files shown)


include/votca/xtp/matrixfreeoperator.h
```diff
@@ -23,8 +23,26 @@
...
+}  // namespace Eigen
+
+namespace votca {
+namespace xtp {
+
+class MatrixFreeOperator : public Eigen::EigenBase<MatrixFreeOperator> {
  public:
+  typedef double Scalar;
+  typedef double RealScalar;
+  typedef int StorageIndex;
```
NicoRenaud: you made me remove that not so long ago

NicoRenaud: do we need them now ?

include/votca/xtp/matrixfreeoperator.h
```diff
@@ -23,8 +23,26 @@
 namespace votca {
 namespace xtp {
 
-class MatrixFreeOperator : public Eigen::EigenBase<Eigen::MatrixXd> {
+class MatrixFreeOperator;
+}
+}  // namespace votca
+namespace Eigen {
+namespace internal {
+// MatrixReplacement looks-like a SparseMatrix, so let's inherits its traits:
```
NicoRenaud: can you remove that comment or at least make it correct ?
It's not a SparseMatrix

include/votca/xtp/matrixfreeoperator.h
```diff
@@ -23,8 +23,26 @@
...
+template <>
+struct traits<votca::xtp::MatrixFreeOperator>
+    : public Eigen::internal::traits<Eigen::MatrixXd> {};
+}  // namespace internal
+}  // namespace Eigen
+
+namespace votca {
+namespace xtp {
+
+class MatrixFreeOperator : public Eigen::EigenBase<MatrixFreeOperator> {
```
NicoRenaud: The change is a bit above my pay grade I'm afraid. But the eigen web page agrees with you.

include/votca/xtp/matrixfreeoperator.h
JensWehner: yes

include/votca/xtp/matrixfreeoperator.h
JensWehner: It is called the Curiously recurring template pattern, basically it is a compile time inheritance MatrixFreeOperator can use the methods of Eigen::Base

include/votca/xtp/matrixfreeoperator.h
JensWehner: nope

PR #304, opened 2019-10-17
Overwrite

JensWehner:
Added overwrite specifiers https://en.cppreference.com/w/cpp/language/override


junghans:
Duplicate of #306

PR #305, opened 2019-10-17
Nullptr

JensWehner:
replace NULL and 0 by nullptr http://www.open-std.org/jtc1/sc22/wg21/docs/papers/2007/n2431.pdf


junghans:
Duplicate of #306

PR #307, opened 2019-10-23
Compute only the thin Q in the QR decompsition

NicoRenaud:
One line that changes a lot the performance

### DIFF
--- a/src/libxtp/davidsonsolver.cc
+++ b/src/libxtp/davidsonsolver.cc
@@ -386,7 +386,7 @@ Eigen::MatrixXd DavidsonSolver::qr(const Eigen::MatrixXd &A) const {
   int nrows = A.rows();
   int ncols = A.cols();
   ncols = std::min(nrows, ncols);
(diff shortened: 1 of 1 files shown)


src/libxtp/davidsonsolver.cc
```diff
@@ -386,10 +386,9 @@ Eigen::MatrixXd DavidsonSolver::qr(const Eigen::MatrixXd &A) const {
   int nrows = A.rows();
   int ncols = A.cols();
   ncols = std::min(nrows, ncols);
-
+  Eigen::MatrixXd I = Eigen::MatrixXd::Identity(nrows,ncols);
   Eigen::HouseholderQR<Eigen::MatrixXd> qr(A);
-  Eigen::MatrixXd result = qr.householderQ();
-  result.conservativeResize(nrows, ncols);
+  Eigen::MatrixXd result = qr.householderQ() * I;
```
JensWehner: you can just return the qr.housholderQ()*I, but both work for me

PR #316, opened 2019-11-12
Davidson pp

JensWehner:
Merge after https://github.com/votca/xtp/pull/315

### DIFF
--- a/include/votca/xtp/davidsonsolver.h
+++ b/include/votca/xtp/davidsonsolver.h
@@ -87,16 +87,16 @@ class DavidsonSolver {
         restart(rep, proj, size_initial_guess);
       } else {
         updateProjection(A, proj, iiter);
-      }
 
-      // get the ritz vectors
-      switch (this->_matrix_type) {
-        case MATRIX_TYPE::SYMM: {
-          rep = getRitz(proj);
-          break;
-        }
-        case MATRIX_TYPE::HAM: {
-          rep = getHarmonicRitz(A, proj);
+        // get the ritz vectors
+        switch (this->_matrix_type) {
+          case MATRIX_TYPE::SYMM: {
+            rep = getRitz(proj);
+            break;
+          }
+          case MATRIX_TYPE::HAM: {
+            rep = getHarmonicRitz(A, proj);
+          }
         }
       }
 
--- a/include/votca/xtp/davidsonsolver.h
+++ b/include/votca/xtp/davidsonsolver.h
@@ -55,12 +55,15 @@ class DavidsonSolver {
   Eigen::VectorXd eigenvalues() const { return this->_eigenvalues; }
   Eigen::MatrixXd eigenvectors() const { return this->_eigenvectors; }
   Eigen::MatrixXd residues() const { return this->_res; }
-  Index num_iterations() const { return this->_num_iter; }
+  Index num_iterations() const { return this->_i_iter; }
 
   template <typename MatrixReplacement>
   void solve(const MatrixReplacement &A, Index neigen,
              Index size_initial_guess = 0) {
 
+    if (_max_search_space < neigen) {
+      _max_search_space = neigen * 5;
+    }
     std::chrono::time_point<std::chrono::system_clock> start =
         std::chrono::system_clock::now();
     Index op_size = A.rows();
@@ -80,13 +83,15 @@ class DavidsonSolver {
     ProjectedSpace proj = initProjectedSpace(neigen, size_initial_guess);
     RitzEigenPair rep;
 
-    for (Index iiter = 0; iiter < _iter_max; iiter++) {
+    for (_i_iter = 0; _i_iter < _iter_max; _i_iter++) {
 
-      // restart or update the projection
-      if (proj.search_space > _max_search_space) {
+      bool do_restart = (proj.search_space() > _max_search_space);
+
+      if (do_restart) {
         restart(rep, proj, size_initial_guess);
+
       } else {
-        updateProjection(A, proj, iiter);
+        updateProjection(A, proj);
 
         // get the ritz vectors
         switch (this->_matrix_type) {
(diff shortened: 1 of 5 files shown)


include/votca/xtp/davidsonsolver.h
```diff
@@ -80,46 +83,43 @@ class DavidsonSolver {
...
+        updateProjection(A, proj);
+
+        // get the ritz vectors
+        switch (this->_matrix_type) {
+          case MATRIX_TYPE::SYMM: {
+            rep = getRitz(proj);
+            break;
+          }
+          case MATRIX_TYPE::HAM: {
+            rep = getHarmonicRitz(A, proj);
```
NicoRenaud: should we break here as well ?

include/votca/xtp/davidsonsolver.h
```diff
@@ -80,46 +83,43 @@ class DavidsonSolver {
...
+          }
         }
-      }
 
-      // etend the subspace
-      Index nupdate = extendProjection(rep, proj);
+        Index nupdate = extendProjection(rep, proj);
+        proj.V = orthogonalize(proj.V, nupdate);
```
NicoRenaud: ha the orth was even if we do not restart ? better here indeed. Should it be directly in extendProjection ?

include/votca/xtp/davidsonsolver.h
```diff
@@ -180,19 +184,16 @@ class DavidsonSolver {
...
       Index old_dim = proj.T.cols();
       Index new_dim = proj.V.cols();
       Index nvec = new_dim - old_dim;
       proj.AV.conservativeResize(Eigen::NoChange, new_dim);
-      proj.AV.block(0, old_dim, size, nvec) =
-          A * proj.V.block(0, old_dim, size, nvec);
-      Eigen::MatrixXd VAV =
-          proj.V.transpose() * proj.AV.block(0, old_dim, size, nvec);
+      proj.AV.rightCols(nvec) = A * proj.V.rightCols(nvec);
+      Eigen::MatrixXd VAV = proj.V.transpose() * proj.AV.rightCols(nvec);
```
NicoRenaud: better with rightCols indeed

src/libxtp/davidsonsolver.cc
```diff
@@ -271,18 +275,17 @@ DavidsonSolver::ProjectedSpace DavidsonSolver::initProjectedSpace(
...
-long DavidsonSolver::extendProjection(DavidsonSolver::RitzEigenPair &rep,
-                                      DavidsonSolver::ProjectedSpace &proj) {
+Index DavidsonSolver::extendProjection(const DavidsonSolver::RitzEigenPair &rep,
+                                       DavidsonSolver::ProjectedSpace &proj) {
 
   Index nupdate = 0;
   for (Index j = 0; j < proj.size_update; j++) {
 
-    // skip the root that have already converged
-    if (this->_matrix_type == MATRIX_TYPE::SYMM) {
```
NicoRenaud: I know I introduced that for "a" reason. It might speed up convergence ... would be nice to test again with it

src/libxtp/davidsonsolver.cc
```diff
@@ -294,11 +297,9 @@ long DavidsonSolver::extendProjection(DavidsonSolver::RitzEigenPair &rep,
     proj.V.rightCols<1>() = w.normalized();
 
     // track converged root
-    proj.root_converged[j] = (rep.res_norm[j] < _tol);
+    proj.root_converged[j] = (rep.res_norm()[j] < _tol);
   }
 
-  proj.search_space = proj.V.cols();
```
NicoRenaud: Nice to repalce that with a method though

src/libxtp/davidsonsolver.cc
```diff
@@ -315,49 +316,40 @@ Eigen::VectorXd DavidsonSolver::computeCorrectionVector(
...
 Eigen::VectorXd DavidsonSolver::dpr(const Eigen::VectorXd &r,
                                     double lambda) const {
-  /* \brief Compute the diagonal preconditoned residue : delta = - (D -
-   * lambda)^{-1} r
+  /* \brief Compute the diagonal preconditoned residue :
+  \delta = -r/(D - lambda)
    */
-  Eigen::VectorXd delta = r.array() / (lambda - _Adiag.array());
-  return delta;
+  return (-r.array() / (_Adiag.array() - lambda));
```
NicoRenaud: why did you rewrite is like that ? Wouldn't it  be marginally more efficient the way it was ?

src/libxtp/davidsonsolver.cc
```diff
@@ -397,6 +385,10 @@ Eigen::MatrixXd DavidsonSolver::gramschmidt(const Eigen::MatrixXd &A,
   Eigen::MatrixXd Q = A;
   for (Index j = nstart; j < A.cols(); ++j) {
     Q.col(j) -= Q.leftCols(j) * (Q.leftCols(j).transpose() * A.col(j));
+    Q.col(j).normalize();
+    // two is enough GS
+    // http://stoppels.blog/posts/orthogonalization-performance
+    Q.col(j) -= Q.leftCols(j) * (Q.leftCols(j).transpose() * Q.col(j));
```
NicoRenaud: did you notice any difference with that ? Cause the numbers were exactly the same when I tried that ... So if it does nothing I would remove the second part. Or at least make it another options (yeah ! another option)

include/votca/xtp/davidsonsolver.h
JensWehner: yeah good point

include/votca/xtp/davidsonsolver.h
JensWehner: Yeah maybe, Probably actually

src/libxtp/davidsonsolver.cc
JensWehner: do you want to do it or should I? I have no idea what the reason was. :)

src/libxtp/davidsonsolver.cc
JensWehner: THis way it is closer to the comment. I do not mind in the end.

JensWehner: The optimised code will look the same most likely.

src/libxtp/davidsonsolver.cc
JensWehner: Actually, yes I have seen a difference in the overlap matrix between the vectors.

PR #333, opened 2019-12-03
State tracker fix

gtirimbo:


JensWehner:
Looking interesting I will refactor it a bit, but we need tests for the new filters.

Issue #336, opened 2019-12-05
share/logo/CH4.cube is 47MB


junghans:
And hence 90% of the release tarballs

JensWehner:
Is compressing it enough?

JensWehner:
makes it 12MB

PR #337, opened 2019-12-05
removed logo from repo

JensWehner:
Fix https://github.com/votca/xtp/issues/336


JensWehner:
@junghans something is wrong with the versions. Help

Issue #338, opened 2019-12-05
unit_test_segment and unit_test_statetracker test fail


junghans:
Testing 1.6_rc1:
```
37/69 Test #37: unit_test_segment .......................***Failed    0.02 sec
Running 2 test cases...
/usr/include/c++/9/bits/stl_vector.h:1042: std::vector<_Tp, _Alloc>::reference std::vector<_Tp, _Alloc>::operator[](std::vector<_Tp, _Alloc>::size_type) [with _Tp = long int; _Alloc = std::allocator<long int>; std::vector<_Tp, _Alloc>::reference = long int&; std::vector<_Tp, _Alloc>::size_type = long unsigned int]: Assertion '__builtin_expect(__n < this->size(), true)' failed.
unknown location(0): [4;31;49mfatal error: in "segment_test/writehdf5": signal: SIGABRT (application abort requested)[0;39;49m
/builddir/build/BUILD/xtp-1.6_rc1/src/tests/test_segment.cc(56): [1;36;49mlast checkpoint: "writehdf5" test entry[0;39;49m
[1;31;49m*** 1 failure is detected in the test module "segment_test"
[0;39;49m
      Start 38: unit_test_aoshell
38/69 Test #38: unit_test_aoshell .......................   Passed    0.02 sec
      Start 39: unit_test_sphere_lebedev_rule
39/69 Test #39: unit_test_sphere_lebedev_rule ...........   Passed    0.02 sec
      Start 40: unit_test_sphere_lebedev_rule_fine
40/69 Test #40: unit_test_sphere_lebedev_rule_fine ......   Passed    0.02 sec
      Start 41: unit_test_sphere_lebedev_rule_xfine
41/69 Test #41: unit_test_sphere_lebedev_rule_xfine .....   Passed    0.02 sec
      Start 42: unit_test_statetracker
42/69 Test #42: unit_test_statetracker ..................***Failed    0.04 sec
Running 2 test cases...
/usr/include/c++/9/bits/stl_vector.h:1042: std::vector<_Tp, _Alloc>::reference std::vector<_Tp, _Alloc>::operator[](std::vector<_Tp, _Alloc>::size_type) [with _Tp = long int; _Alloc = std::allocator<long int>; std::vector<_Tp, _Alloc>::reference = long int&; std::vector<_Tp, _Alloc>::size_type = long unsigned int]: Assertion '__builtin_expect(__n < this->size(), true)' failed.
[...]

PR #345, opened 2019-12-09
do not read empty vectors

JensWehner:
fixes  https://github.com/votca/xtp/issues/338


junghans:
Testing here: https://koji.fedoraproject.org/koji/taskinfo?taskID=39466902

PR #359, opened 2020-01-06
forgot to add external energy to total energy

JensWehner:
nuclei external charge interaction energy was not added to total energy

if external charge sits on a nucleus, contribution is ignored but warnings is logged.


baumeier:
Can we output that energy contribution (also) separately? The nuclear repulsion energy alone is printed in the beginning. Would also benefit from higher precision.

PR #361, opened 2020-01-07
fixes

JensWehner:
Cleaned up davidson a bit and 
simplified some matrixmultiplication.


JensWehner:
yeah that was bad....

JoshuaSBrown:
hahahaha, that's awesome.

PR #366, opened 2020-01-13
renamed register filter

JensWehner:


gtirimbo:
Changing from ¨T¨ to ¨t¨ solve the issue

PR #371, opened 2020-01-16
Better QP convergence and ppm update + benchmark

JensWehner:
- ppm model replace the cos screening with x/x2+eta2 screening **changes results**
- implemented frequency derivatives of sigma
- replaced fixed point by newton iterations
- new newton iterator class
- changed tests
- added bisection method to grid search
- changed selection criterium of roots
- added a Thiel set benchmark suite to share


include/votca/xtp/IndexParser.h
```diff
@@ -0,0 +1,45 @@
...
+ */
+
+#pragma once
+#ifndef VOTCA_XTP_INDEXPARSER_H
+#define VOTCA_XTP_INDEXPARSER_H
+
+#include <votca/tools/types.h>
+
+/**
+ * \brief Parser to read strings containing indeces in the format "1 2...7 8"
```
JavierSijen: indeces -> indices/indexes

include/votca/xtp/IndexParser.h
```diff
@@ -0,0 +1,45 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ */
+
+#pragma once
+#ifndef VOTCA_XTP_INDEXPARSER_H
+#define VOTCA_XTP_INDEXPARSER_H
+
```
JavierSijen: #include <string>
#include <vector>

src/libxtp/gwbse/gw.cc
```diff
@@ -203,83 +194,146 @@ void GW::CalculateGWPerturbation() {
...
+  return _Sigma_x.diagonal() + _Sigma_c.diagonal() - _vxc.diagonal() +
+         _dft_energies.segment(_opt.qpmin, _qptotal);
+}
+
+Eigen::VectorXd GW::SolveQP(const Eigen::VectorXd& frequencies) const {
+  const Eigen::VectorXd intercepts =
       _dft_energies.segment(_opt.qpmin, _qptotal) + _Sigma_x.diagonal() -
       _vxc.diagonal();
   Eigen::VectorXd frequencies_new = frequencies;
+  Eigen::ArrayXi converged = Eigen::ArrayXi::Zero(_qptotal);
```
JavierSijen: Why not a vector of booleans? Eigen has nice reduction methods like .any() and .all(), which you could use later to check whether all states have converged.

src/libxtp/gwbse/gw.cc
```diff
@@ -203,83 +194,146 @@ void GW::CalculateGWPerturbation() {
...
-        break;  // Converged
-      } else if (_opt.g_sc_max_iterations != 0 &&
-                 g_iter == _opt.g_sc_max_iterations - 1) {
-        break;  // Not converged
+  std::pair<double, double> temp =
+      _sigma->CalcCorrelationDiagElement(gw_level, frequency0);
+  double sigma_x_m_Vxc = intercept0 - frequency0;
+  double Z = temp.second + 1.0;
+  if (std::abs(Z) > 1e-9) {
+    newf = frequency0 + (sigma_x_m_Vxc + temp.first) / Z;
```
JavierSijen: Hmmm isn't it
```
newf = frequency0 - (sigma_x_m_Vxc + temp.first) / Z;
```
i.e. a minus sign instead of plus?

src/libxtp/gwbse/sigma_exact.cc
```diff
@@ -35,26 +35,31 @@ void Sigma_Exact::PrepareScreening() {
...
     temp.segment(n_occ, n_unocc) -= eigenvalue;
-    const Eigen::ArrayXd numer = res_12.array() * temp;
-    const Eigen::ArrayXd denom = temp.abs2() + eta * eta;
-    sigma_c += (numer / denom).sum();
+    const Eigen::ArrayXd numer = res_12 * temp;
+    const Eigen::ArrayXd denom = temp.abs2() + eta2;
+    double sigma_temp = (numer / denom).sum();
+    sigma.first += sigma_temp;
+    sigma.second +=
+        eta2 * (res_12 / denom.abs2()).sum() - std::pow(sigma_temp, 2);
```
JavierSijen: Pretty sure it should be:
```
sigma.second -= (res_12 * (temp.abs2() - eta2) / denom.abs2()).sum();
```

JensWehner: is that not the same?

JensWehner: lala

src/libxtp/gwbse/gw.cc
JensWehner: Good point.

JensWehner:
@JavierSijen please review
@baumeier please acknowledge, this will change results slightly, the benchmark might be useful for further stories, so not everybody has to write a new framework to run the thiel set.

src/libxtp/gwbse/sigma_exact.cc
```diff
@@ -97,13 +97,12 @@ Eigen::MatrixXd Sigma_Exact::CalcResidues(Index gw_level) const {
...
   vc2index vc = vc2index(0, 0, n_unocc);
-  const Eigen::MatrixXd Mmn_i_T = _Mmn[gw_level + qpoffset].transpose();
+  const Eigen::MatrixXd& Mmn_i = _Mmn[gw_level + qpoffset];
   Eigen::MatrixXd res = Eigen::MatrixXd::Zero(_rpatotal, rpasize);
   for (Index v = 0; v < n_occ; v++) {  // Sum over v
-    const Eigen::MatrixXd& Mmn_v = _Mmn[v].block(n_occ, 0, n_unocc, auxsize);
-    const Eigen::MatrixXd fc = Mmn_v * Mmn_i_T;  // Sum over chi
-    const Eigen::MatrixXd& XpY_v =
-        _rpa_solution.XpY.block(vc.I(v, 0), 0, n_unocc, rpasize);
+    auto Mmn_v = _Mmn[v].block(n_occ, 0, n_unocc, auxsize);
```
JavierSijen: Why use auto keyword? Does it prevent Eigen from casting their result into a matrix?

src/libxtp/gwbse/gw.cc
```diff
@@ -328,18 +324,21 @@ double GW::SolveQP_Bisection(double lowerbound, double f_lowerbound,
...
-      break;
       zero = c;
+      break;
     }
     std::pair<double, double> temp = f(c);
     double y_c = temp.first;
     if (std::abs(y_c) < _opt.g_sc_limit) {
-      break;
       zero = c;
+      break;
```
JavierSijen: If this breaks the while-loop, in most cases, it would've broken at the next iteration anyways. So this break might be slightly inefficient?

src/libxtp/gwbse/gw.cc
```diff
@@ -235,7 +235,7 @@ Eigen::VectorXd GW::SolveQP(const Eigen::VectorXd& frequencies) const {
     }
   }
 
-  if ((converged == true).all()) {
+  if ((converged == false).any()) {
```
JavierSijen: Couldn't it be:
`!converged.all()`

src/libxtp/gwbse/gw.cc
```diff
@@ -328,18 +324,21 @@ double GW::SolveQP_Bisection(double lowerbound, double f_lowerbound,
   while (true) {
     double c = 0.5 * (lowerbound + upperbound);
     if (std::abs(upperbound - lowerbound) < _opt.g_sc_limit) {
```
JavierSijen: In the first if statement, the x-value is compared to the tolerance. In the second if statement, the y-value is compared to the tolerance. So, inconsistent units for `_opt.g_sc_limit`!

src/libxtp/gwbse/sigma_exact.cc
JensWehner: yeah it is just an expression, I do not want a copy.

src/libxtp/gwbse/gw.cc
```diff
@@ -270,31 +270,27 @@ boost::optional<double> GW::SolveQP_Grid(double intercept0, double frequency0,
...
-      std::pair<double, double> temp3 =
-          _sigma->CalcCorrelationDiagElement(gw_level, f);
-      double pole_weight = 1.0 / (1.0 - temp3.second);
-      if (pole_weight >= 1e-5 && pole_weight > pole_weight_max) {
+    std::pair<double, double> targ = fqp(freq);
+    if (targ_prev.first * targ.first < 0.0) {  // Sign change
+      double f =
+          SolveQP_Bisection(freq_prev, targ_prev.first, freq, targ.first, fqp);
+      std::pair<double, double> temp3 = fqp(f);
+      double pole_weight = -1.0 / temp3.second;
```
JavierSijen: Is this the only usage for the derivative of sigma_c? If so, it might be more efficient so simply use a first order approximation of the derivative. But then all the sigma_c derivative code is wasted...

src/libxtp/gwbse/gw.cc
JensWehner: wikipedia des it the same way and we only want the frequency to be accurate to `_opt.g_sc_limit`

src/libxtp/gwbse/gw.cc
JensWehner: it is slightly more efficient, I just translated the wikipedia pseudo code

src/libxtp/gwbse/gw.cc
JensWehner: The newton iteration stuff still uses it, but it is also split up now into two functions.

src/libxtp/gwbse/gw.cc
JensWehner: good question. I am not sure and a bit lazy to try now.

JensWehner: you are right and it looks much cleaner.

JensWehner: thanks

PR #378, opened 2020-02-07
RPA call in BSE Operator uses KS energies for G0W0

baumeier:


baumeier:
will be done in a different one

src/libxtp/gwbse/bse.cc
```diff
@@ -49,7 +49,11 @@ void BSE::configure(const options& opt, const Eigen::VectorXd& DFTenergies) {
 void BSE::SetupDirectInteractionOperator(const Eigen::VectorXd& DFTenergies) {
   RPA rpa = RPA(_log, _Mmn);
   rpa.configure(_opt.homo, _opt.rpamin, _opt.rpamax);
-  rpa.UpdateRPAInputEnergies(DFTenergies, _Hqp.diagonal(), _opt.qpmin);
+  if ( isG0W0 ){ 
+    rpa.setRPAInputEnergies(DFTenergies.segment(_opt.rpamin, _opt.rpamax - _opt.rpamin + 1));
```
JensWehner: there can also be a shift from the scissor effect, so this is not entirely correct.

PR #382, opened 2020-02-14
Fix rpa indexing

JavierSijen:
RPA::UpdateRPAInputEnergies remakes the RPA energies vector `_energies` of size 1 x rpatotal. If I understand it correctly, RPA::UpdateRPAInputEnergies is supposed to do three things:
1. Initialize `_energies` with the DFT energies: `_energies = dftenergies(rpamin:rpamax)`
2. Fill the QP range with the latest GW energies: `_energies(qpmin:qpmax) = gwaenergies`
3. Shift the trailing DFT energies such that they have the same shift (from HOMO-LUMO gap) as the GW energies: `_energies(qpmax+1:end) += (QPGap - DFTGap)`

The problem is: In master, `_energies` is indexed using a `-_rpamin` offset, as if `_energies` is the same size as `dftenergies`. But `_energies` is of size 1 x rpatotal. So indexing of `_energies` should not be offset with `-_rpamin`!

It looks as though `_energies` was at one point the same size as `dftenergies`.

TODO: Is the same indexing mistake in other places?

### DIFF
--- a/src/libxtp/gwbse/rpa.cc
+++ b/src/libxtp/gwbse/rpa.cc
@@ -28,18 +28,16 @@ namespace xtp {
 void RPA::UpdateRPAInputEnergies(const Eigen::VectorXd& dftenergies,
                                  const Eigen::VectorXd& gwaenergies,
                                  Index qpmin) {
-  Index rpatotal = _rpamax - _rpamin + 1;
+  const Index rpatotal = _rpamax - _rpamin + 1;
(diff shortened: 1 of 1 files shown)


JensWehner:
are you sure this is correct?

JavierSijen:
> 
> 
> are you sure this is correct?

Whoops, I was wrong. I looked at it again. This diagram explains why the original code is correct:
rpa_indexing

Nothing to see here.

PR #386, opened 2020-02-21
removed duplicate basis set files

### LINKED ISSUE #384: Duplicated basissets
On a case-insensitive file-system git cannot checkout `aux-cc-pVTZ.xml`, `aux-def2-SVP.xml`, `cc-pVTZ.xml` and `def2-SVP.xml` as filename with different capitalization exist, e.g. `aux-cc-pVTZ.xml` and `aux-aug-cc-pvtz.xml`.

baumeier:
Fixes #384


junghans:
This should have gone into stable.

Issue #388, opened 2020-02-23
Extend BSE active space to include levels < qpmin and > qpmax


baumeier:
In some cases, it would be useful to restrict accurate GW corrections to a smaller set of molecular orbitals and include appropriately shifted levels below qpmin and above qpmax in the construction of the BSE.
This can be useful in partiuclar for slowly or hardly converging evGW calculation on the full set of molecular orbitals, and is done in similar fashion in the literature.
@JensWehner pointers to where changes are needed to achieve this are appreciated.

JensWehner:
It is all in the bse_ooerator.cc hqp Part. So you have to extend the quasiparticle hamiltonian and remove the checks in gwbse.cc

PR #390, opened 2020-02-25
Bse extension

baumeier:
Implements #388 
- Choice of level range in the BSE is now decoupled from QP range.
- if BSEmin < QPmin or BSEmax > QPmax, shifted KS energies are put on the diagonal of Hqp
- padding or cutting is now done before BSE object is created, construction of Hqp in bse_operator adjusted accordingly


src/libxtp/gwbse/bse_operator.cc
```diff
@@ -94,11 +94,11 @@ Eigen::RowVectorXd BSE_OPERATOR<cqp, cx, cd, cd2>::Hqp_row(Index index) const {
   Index v1 = vc.v(index);
   Index c1 = vc.c(index);
   Eigen::MatrixXd Result = Eigen::MatrixXd::Zero(_bse_ctotal, _bse_vtotal);
-  Index cmin = _bse_cmin - _opt.qpmin;
```
JensWehner: is this general?

src/libxtp/gwbse/bse_operator.cc
```diff
@@ -94,11 +94,11 @@ Eigen::RowVectorXd BSE_OPERATOR<cqp, cx, cd, cd2>::Hqp_row(Index index) const {
...
+  Index cmin = _bse_vtotal;
   // v->c
   Result.col(v1) = _Hqp.col(c1 + cmin).segment(cmin, _bse_ctotal);
   // c-> v
-  Index vmin = _opt.vmin - _opt.qpmin;
+  Index vmin = 0;  //_opt.vmin - _opt.qpmin;
```
JensWehner: why zero?

src/libxtp/gwbse/gwbse.cc
```diff
@@ -159,12 +159,12 @@ void GWBSE::Initialize(tools::Property& options) {
   }
 
   // some QP - BSE consistency checks are required
-  if (bse_vmin < qpmin) {
+  /* if (bse_vmin < qpmin) {
```
JensWehner: please remove commented out code

src/libxtp/gwbse/gwbse.cc
```diff
@@ -651,7 +652,7 @@ bool GWBSE::Evaluate() {
       << flush;
 
   Eigen::MatrixXd Hqp;
```
JensWehner: scope of this variable can be reduced.

src/libxtp/gwbse/gwbse.cc
```diff
@@ -721,5 +724,60 @@ bool GWBSE::Evaluate() {
       << TimeStamp() << " GWBSE calculation finished " << flush;
   return true;
 }
+
+Eigen::MatrixXd GWBSE::PrepareHQP_forBSE(const Eigen::MatrixXd& Hqp) {
```
JensWehner: needs a better name than Prepare, maybe Scale or Expand or so. Prepare is pretty generic.

src/libxtp/gwbse/gwbse.cc
```diff
@@ -721,5 +724,60 @@ bool GWBSE::Evaluate() {
...
+  Index RPAoffset = _bseopt.vmin - _gwopt.rpamin;
+  Eigen::MatrixXd Hqp_BSE = Eigen::MatrixXd::Zero(hqp_size, hqp_size);
+
+  // if bse range is smaller than qprange, cut out  Hqp block
+  if (_bseopt.vmin >= _gwopt.qpmin) {
+    Index start = _bseopt.vmin - _gwopt.qpmin;
+    if (_bseopt.cmax <= _gwopt.qpmax) {
+      Hqp_BSE = Hqp.block(start, start, hqp_size, hqp_size);
+    } else {
+      Hqp_BSE.block(0, 0, gwsize - start, gwsize - start) =
```
JensWehner: `gwsize - start ` should get its own name.

JensWehner: instead of `block` use `topLeftCorner` when you start at 0,0

src/libxtp/gwbse/gwbse.cc
```diff
@@ -721,5 +724,60 @@ bool GWBSE::Evaluate() {
...
+          Hqp.block(start, start, gwsize - start, gwsize - start);
+
+      // now extend
+      Index virtoffset = gwsize - start;
+      Index virt_extra = _bseopt.cmax - _gwopt.qpmax;
+      for (Index i_virt = 0; i_virt < virt_extra; ++i_virt) {
+        Hqp_BSE(virtoffset + i_virt, virtoffset + i_virt) =
```
JensWehner: `Hqp_BSE.diagonal().tail(virt_extra)=_orbitals.RPAInputEnergies().segment(RPAoffset + virtoffset,virt_extra)` 

instead of the for loop

src/libxtp/gwbse/gwbse.cc
```diff
@@ -721,5 +724,60 @@ bool GWBSE::Evaluate() {
...
+            _orbitals.RPAInputEnergies()(RPAoffset + virtoffset + i_virt);
+      }
+    }
+  }
+
+  if (_bseopt.vmin < _gwopt.qpmin) {
+    Index occ_extra = _gwopt.qpmin - _bseopt.vmin;
+    // put RPA input energies in top left corner _bsemin < _qpmin
+    for (Index i_occ = 0; i_occ < occ_extra; ++i_occ) {
+      Hqp_BSE(i_occ, i_occ) = _orbitals.RPAInputEnergies()(RPAoffset + i_occ);
```
JensWehner: this is also an eigen call.

src/libxtp/gwbse/gwbse.cc
```diff
@@ -721,5 +724,60 @@ bool GWBSE::Evaluate() {
...
+    }
+    // put explicit Hqp on _qpmin:_qpmax block
+    Hqp_BSE.block(occ_extra, occ_extra, gwsize, gwsize) = Hqp;
+
+    if (_bseopt.cmax > _gwopt.qpmax) {
+
+      // now extend
+      Index virtoffset = occ_extra + gwsize;
+      Index virt_extra = _bseopt.cmax - _gwopt.qpmax;
```
JensWehner: Is this basically the same as above? Use eigen calls instead of for loop, please.

baumeier:
@JensWehner Test of the new HQP modification function in the GWBSE class? If so, any suggestion how? Is it ok to have a completely fake matrix that is being extended by entries from a fake energy vector? Otherwise, it seem we will need to almost put the whole GW steps in the test.

src/libxtp/gwbse/bse_operator.cc
baumeier: The HQP matrix that is used to initialize the BSE object now should exactly contain only the entries from vmin:cmax.

JensWehner: Can you add the comment then please.

JensWehner:
actually, is should be in the bse class not in the GW class, because it has nothing todo with GW, but only with BSE. So basically you input a Hqp from GW and check that the output is in the BSE format you expect.

baumeier:
Are you talking now about the function itself?

src/libxtp/gwbse/bse_operator.cc
baumeier: Where?

baumeier:
OK, I think one can also make the function a member of BSE class...

JensWehner:
both.

src/libxtp/gwbse/bse.cc
```diff
@@ -44,9 +44,51 @@ void BSE::configure(const options& opt,
   _bse_vtotal = _bse_vmax - _opt.vmin + 1;
   _bse_ctotal = _opt.cmax - _bse_cmin + 1;
   _bse_size = _bse_vtotal * _bse_ctotal;
+  _Hqp = Scale_or_Expand_HQP(RPAInputEnergies);
   SetupDirectInteractionOperator(RPAInputEnergies);
 }
 
+Eigen::MatrixXd BSE::Scale_or_Expand_HQP(
```
JensWehner: Looks much nicer but I do not get the scale part of the name?

baumeier: How about Cut_or_Expand?
Actually it will do both, if required at the same time. Like if bsemin > qpmin but bsemax > qpmax, it will "cut" for min and expand for max. Maybe CutExpand?

JensWehner: AdjustHqpSize?

baumeier:
@JensWehner ping

src/libxtp/gwbse/bse.cc
baumeier: Done

src/libxtp/gwbse/gwbse.cc
baumeier: Done

src/libxtp/gwbse/gwbse.cc
baumeier: Done

src/libxtp/gwbse/gwbse.cc
baumeier: Done in new BSE function

src/libxtp/gwbse/gwbse.cc
baumeier: Done

baumeier: Done.

src/libxtp/gwbse/gwbse.cc
baumeier: Done

JensWehner:
I am not super happy with the test yet. It is really bad practice to add a method just to test something. I have no better idea though atm.

baumeier:
Okay, anything else?

JensWehner:
It is probably a good idea to remove the Hqp from the constructor and pass it in the configure method, because then we do not have to keep both Hqp and Hqp_expanded as member variables.

baumeier:
Changed.

PR #393, opened 2020-02-27
xml and logger update

rubengerritsen:
While using votca I encountered two "problems". 

The first, in the `iqm.cc` file, a log file for the dft calculation is only created if the dft calculation is successful. On failure, however, no `dft.log` is printed. I simply added a line of code that prints a `dft.log` on failure.

The second, in the `xtpdft.xml` file, `max_iterations` was placed in the `package` element, but the code tries to read it from the `package/convergence` element. I simply moved it so the code can read it. (for how the option is read see: `dftengine.cc` line 114)


JensWehner:
sounds good.

PR #401, opened 2020-03-06
check executable for dft jobs other than xtp

### LINKED ISSUE #400: Check path to executable to executable only if dftpackage != xtp
Currently, we are checking that there must be an `executable` path for all the dft jobs including xtp

felipeZ:
fixed #400


felipeZ:
@JensWehner @baumeier  The idea is to check the executable's path only for the qmpackages other than xtp

Issue #402, opened 2020-03-07
Option to use offdiagonal elements of Hqp in BSE


baumeier:
Including the symmetrized offdiagonal elements of the quasiparticle Hamiltonian in the BSE can lead to spurious effects, such as discontinuities in excitation energies as a function of bond length in diatomic molecules. 
Use of these offdiagonal should be optional.

JensWehner:
my gut feeling: 

when we use Hqp_diag we are basically doing everything in the MO basis

if we use the offdiag of Hqp we could also just use the eigenvalues and eigenvectors of Hqp to build H_BSE, how would we have to modify H_d and H_x then?

baumeier:
There is no reference, it is an observation. The underlying issue is the unresolved question how to deal with the fact that including the offdiagonal elements makes Hqp not symmetric anymore. AFAIK the other codes ignore it, there is only https://pubs.acs.org/doi/10.1021/acs.jctc.5b00394 in which it has been explored recently (and where the effects where deemed to be small compared to an approximate update of the eigenvalues). In Fig.2 one can clearly see that Hqp is not symmetric, and eigenvalues are not guaranteed to be real and that should cause havoc in the current way the BSE is implemented. 

For the moment, I would like to have the option to ignore the offdiagonals or take them into account in the symmetric form as we do now. The rest takes a major investigation, I think. PR incoming.

PR #403, opened 2020-03-07
Use of off-diagonal elements of Hqp in BSE optional

baumeier:
implements #402

### DIFF
--- a/include/votca/xtp/bse.h
+++ b/include/votca/xtp/bse.h
@@ -62,6 +62,7 @@ class BSE {
     Index davidson_maxiter = 50;
     double min_print_weight =
         0.5;  // minimium contribution for state to print it
+    bool use_Hqp_offdiag = false;
   };
 
   void configure(const options& opt, const Eigen::VectorXd& RPAEnergies,
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,7 +6,8 @@ For more detailed information about the changes see the history of the
 * Hqp part of BSE can be separate from QP range (#388)
 * Fixed Cuda implementation (#391)
 * Use a general syntax to represent all QMPackages input (#318)
-* Remove support for both Gaussian and NWChem (#318)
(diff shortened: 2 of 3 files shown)


JensWehner:
If you changed the defaults, why did all the tests pass?

baumeier:
Because the option only changes which Hqp (full or diagonal only) is passed to the configure function of the BSE. The Hqp in the test is unchanged and no method is changed.

include/votca/xtp/bse.h
```diff
@@ -62,6 +62,7 @@ class BSE {
     Index davidson_maxiter = 50;
     double min_print_weight =
         0.5;  // minimium contribution for state to print it
+    bool use_Hqp_offdiag = false;
```
JensWehner: What does this line do?

JensWehner: okay gwbse.cc has no tests so far, but then please remove this line.

baumeier: And where else is it supposed to go instead?

JensWehner: where do you use it?

JensWehner: I cannot find any place it is needed.

baumeier: In gwbse.cc only. Maybe it is better instead of changing the call to bse.configure in gwbse.cc depending on the option, to have the ignoring/including of the offdiags in the configure function, by

```suggestion
if ( _opt.use_Hqp_offdiag ){
_Hqp = AdjustHqpSize(Hqp_in, RPAInputEnergies);
} else {
_Hqp = AdjustHqpSize(Hqp_in, RPAInputEnergies).diagonal().asDiagonal();
}
```

baumeier: Different files, so the change is to be in bse.cc

JensWehner: yes that is a better approach. Good idea.

PR #406, opened 2020-03-09
Refactor of aobasis

JensWehner:
- Removed all the recursion in basisset.cc and replaced it with for loops for better readability and added some tests
- Moved ReorderMOs from AOBasis to QMPackage
- Input format of types is now checked
- The Reorder parameters are now stored in the qmpackage and not hardcoded in aobasis.cc

### DIFF
--- a/include/votca/xtp/qmpackage.h
+++ b/include/votca/xtp/qmpackage.h
@@ -118,6 +118,16 @@ class QMPackage {
   std::vector<std::string> GetLineAndSplit(std::ifstream& input_file,
                                            const std::string separators) const;
 
+  // each qmpackage has its own ordering of the individual functions in each
+  // shell
+  // i.e. VOTCA uses z,y,x e.g. Y1,0 Y1,-1 Y1,1 for the p shell
+  // d3z2-r2 dyz dxz dxy dx2-y2 e.g. Y2,0 Y2,-1 Y2,1 Y2,-2 for the d shell and
+  // so forth. ORCA uses z,x,y for the p shell
+  // these methods reorder the MOs to that format using the
+  // ShellReorder() and ShellMulitplier() which specify the order for each
+  // QMPackage. Some codes also use different normalisation conditions which
+  // lead to other signs for some of the entries, which can be changed via the
+  // multipliers.
   void ReorderMOsToXTP(Eigen::MatrixXd& v, const AOBasis& basis) const;
   void ReorderMOsToNative(Eigen::MatrixXd& v, const AOBasis& basis) const;
 
--- a/include/votca/xtp/qmpackage.h
+++ b/include/votca/xtp/qmpackage.h
@@ -1,5 +1,5 @@
 /*
- *            Copyright 2009-2019 The VOTCA Development Team
+ *            Copyright 2009-2020 The VOTCA Development Team
  *                       (http://www.votca.org)
  *
  *      Licensed under the Apache License, Version 2.0 (the "License")
@@ -18,8 +18,8 @@
  */
 
 #pragma once
-#ifndef VOTCA_XTP_QM_PACKAGE_H
-#define VOTCA_XTP_QM_PACKAGE_H
+#ifndef VOTCA_XTP_QMPACKAGE_H
+#define VOTCA_XTP_QMPACKAGE_H
 
 #include "votca/xtp/aobasis.h"
 #include <votca/tools/property.h>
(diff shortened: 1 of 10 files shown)


src/libxtp/basisset.cc
```diff
@@ -23,136 +23,100 @@ namespace votca {
...
-      lmax = 4;
-    } else if (type == "H") {
-      lmax = 5;
-    } else if (type == "I") {
-      lmax = 6;
-    } else {
-      throw std::runtime_error("FindLmax: Shelltype not known");
-    }
+
+  assert(!type.empty() && "Shelltype most be non empty!");
```
baumeier: “must” :)

src/libxtp/basisset.cc
```diff
@@ -23,136 +23,100 @@ namespace votca {
...
+  } else if (t == 'I') {
+    lmax = 6;
   } else {
-    for (Index i = 0; i < Index(type.length()); ++i) {
-      std::string local_shell = std::string(type, i, 1);
-      Index test = FindLmax(local_shell);
-      if (test > lmax) {
-        lmax = test;
-      }
-    }
```
baumeier: What will happen if there is a SP shell or similar in the basis?

JensWehner: it always takes the last character of the shell. These routines now assume that a) the shell characters are sorted and continous.

baumeier: Makes sense.

src/libxtp/basisset.cc
```diff
@@ -23,136 +23,100 @@ namespace votca {
...
-      std::string local_shell = std::string(type, i, 1);
-      Index test = FindLmin(local_shell);
-      if (test == 0) {
-        return 0;
-      }
-      if (test < lmin) {
-        lmin = test;
-      }
-    }
+    throw std::runtime_error("FindLmax: Shelltype not known");
```
baumeier: FindLmin

JensWehner:
I am considering doing away with all the characters for shells and just have an lmin and lmax. WHat do you thing?

JensWehner:
@felipeZ please review as well.

include/votca/xtp/qmpackage.h
```diff
@@ -118,6 +118,23 @@ class QMPackage {
   std::vector<std::string> GetLineAndSplit(std::ifstream& input_file,
                                            const std::string separators) const;
 
+  void ReorderMOsToXTP(Eigen::MatrixXd& v, const AOBasis& basis) const;
```
felipeZ: I still think that for people new to the code a brief comment explaning what those methods do would be very useful :)

src/libxtp/qmpackages/orca.h
```diff
@@ -57,7 +57,29 @@ class Orca : public QMPackage {
...
 
+ protected:
+  const std::array<Index, 25>& ShellMulitplier() const final {
+    return _multipliers;
+  }
+  const std::array<Index, 25>& ShellReorder() const final { return _reorder; }
+
  private:
+  // clang-format off
+  std::array<Index,25> _multipliers={
```
felipeZ: Could you please explain to me what are those multipliers?

src/libxtp/basisset.cc
```diff
@@ -174,70 +138,75 @@ std::vector<Index> NumFuncSubShell(const std::string& shell_type) {
...
-    } else {
-      throw std::runtime_error(
-          "OffsetFuncShell_cartesian shell_type not known");
-    }
+bool CheckShellType(const std::string& shelltype) {
+  if (shelltype.empty()) {
+    return false;
+  }
+  std::vector<char> allowed_shells = {'S', 'P', 'D', 'F', 'G', 'H', 'I'};
+  std::vector<char>::iterator it =
```
felipeZ: **auto** can help here to reduce the *Carpal tunnel syndrome* ;)

src/libxtp/qmpackage.cc
```diff
@@ -140,5 +138,118 @@ std::string QMPackage::FindDefaultsFile() const {
   return xmlFile;
 }
 
+void QMPackage::ReorderMOs(Eigen::MatrixXd& v,
```
felipeZ: The methods from here to end of the file are some much better and cleaner! Awesome work :+1:

src/libxtp/basisset.cc
JensWehner: :) I prefer the explicit statements a bit more to be honest. I see your point though.

src/libxtp/qmpackages/orca.h
JensWehner: to be honest I do not really get it myself. In my mind they should not be there but @baumeier found that without them it does not work. @baumeier want to explain?

baumeier: Not much to explain. At some point, when we compared the coefficients coming out of orca to other codes (and our own), some entries associated to these functions showed a minus sign difference. BUT only on some elements, not the whole vector. 
Orca is using libint, right? Maybe one can look into that to see if there is a difference in definition of the affected AOs? @felipeZ don‘t you have some experience with libint?

JensWehner: it could actually be a different normalisation condition.

baumeier: Also possible, but orca is a bit too black boxy overall. I don‘t think they have these details documented.

baumeier: Nor do we...

JensWehner: it is in the code :)

JensWehner: ...somewhere

felipeZ: @baumeier  I have experience with libint2, I use it to compute multipole integrals using CP2K optimized basissets

PR #410, opened 2020-03-23
Reorder fix

JensWehner:
- fixes reorder and multiplier bugs.
- no test yet, will be done in separate PR


felipeZ:
@JensWehner, just to satisfy my curiosity :), in this PR you are still using the XTP orbitals order? meaning that you are still not enforcing Orca or Libint ordering. You told me that you would like to use either of them. The `_reorder` array for XTP is now full of zeros instead of ones.

JensWehner:
basically, the reorder array indicates how functions have to be shifted. In the case of xtpdft no function has to be shifted so it is all zeros now.

JensWehner:
it is still the votca ordering.

src/libxtp/qmpackages/xtpdft.h
```diff
@@ -71,10 +71,22 @@ class XTPDFT : public QMPackage {
   const std::array<Index, 25>& ShellReorder() const final { return _reorder; }
 
  private:
-  std::array<Index, 25> _multipliers = {1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1,
-                                        1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1};
-  std::array<Index, 25> _reorder = {1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1,
-                                    1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1};
+  // clang-format off
+  std::array<Index,25> _multipliers={
```
felipeZ: This layout greatly improves the readability :+1:

src/libxtp/qmpackage.cc
```diff
@@ -233,10 +230,11 @@ std::vector<Index> QMPackage::getReorderVector(const AOBasis& basis) const {
 }
 
 std::vector<Index> QMPackage::getReorderShell(const AOShell& shell) const {
-  // multipliers were all found using code, hard to establish
   std::vector<Index> reorder;
   for (Index i = 0; i < shell.getNumFunc(); i++) {
-    reorder.push_back(ShellReorder()[shell.getOffset() + i]);
+    // Shellreorder tells how much a certain funktion has to be shifted with
```
felipeZ: This comment is pure gold :+1:

src/libxtp/qmpackage.cc
```diff
@@ -164,7 +163,6 @@ void QMPackage::ReorderMOsToXTP(Eigen::MatrixXd& v,
                                 const AOBasis& basis) const {
 
   std::vector<Index> multiplier = getMultiplierVector(basis);
-  MultiplyMOs(v, multiplier);
```
felipeZ: It is unclear to my what were doing those `multiplier`s before and why we don't need them anymore.

JensWehner: We only have to multiply once in the conversion process, I understood the old routine which did the back and to transformations together incorrectly. I think the multipliers come because of different normalization conditions for the regular harmonics, but I genuinely do not know. Bjoern introduced them because otherwise it would not work.

PR #413, opened 2020-03-31
Reworked Default/Optional XML 

### LINKED ISSUE #319: Add Default values to xml
Add back to the XML files in the share folder the default values hardcoded in the code

baumeier:
Very first attempt at getting new default xml file behavior, testing on dftgwbse calculator:

- full dftgwbse.xml file in share folder
- calculator passes dftpackage options and gwbse options explicitly through, instead of the xml files
- xml file reading only on calculator level

Still not good:
- basisset, functional definitions are repeating
- some tags are superfluous
- xtp_tools still needs to be passed an user option file on the command line, which can be empty though

Fix #319


JensWehner:
I am not sure if all options should have a default, i.e. xyz file. Why do you want to default everything? I mean we can also put the xyz input into the xml file, if we want to reduce the number of files

baumeier:
The xyz input default does not hurt anyone. Adding the coordinates into the xml file is something I don't like. What people will always have is a coordinate file in most likely that format. Asking them to add xml tags around it is just going to be annoying for people.

And why defaulting everything? Just consider the minimum input file needed for `orca` 
`* xyzfile 0 1 system.xyz
`
It will do HF with def2-SVP.

JensWehner:
so orca does not default on the `system.xyz` as votca will do now.

baumeier:
Yes. We can "undefault" the xyz file name of course since a calculator as defined way down in `tools` or rather the `application` that `xtpapplicaton` is derived from strictly requires an optionfile to be supplied. Now in the full default case that would be a ridiculously looking thing like 
```
<options>
<dftgwbse/>
</options>
```
If we require xyzfile to be specified, at least it will make sense to require the options file:
```<options>
<dftgwbse>
<molecule>system.xyz></molecule>
</dftgwbse>
</options>`

**Comment by junghans** [comment]
Merge after https://github.com/votca/votca/pull/263

File: CHANGELOG.md [hunk 1]
```diff
@@ -11,6 +11,7 @@ For more detailed information about the changes see the history of the
 * Fixed executable path check (#400)
 * Usage of offdiagonal elements of Hqp in BSE optional, default: with offdiagonals (#402)
 * refactored MO reordering of external QMPackages
+* Add defaults to XML (#319)
```
junghans, line 14: Add #413 here.

junghans:
It makes one test fail:
```
 230/244 Test #230: unit_test_orca .....................................***Failed    0.05 sec
 Running 5 test cases...
 Warning: VOTCASHARE environment variable not defined
 /builds/votca/votca/votca/xtp/src/tests/test_orca.cc(8634): error: in "orca_test/input_generation_version_4_0_1": check inp.substr(index1, index2 - index1) == "%basis\nGTOName =\"system.bas\";\nGTOAuxName =\"system.aux\";\n" has failed [%basis
 GTOName ="system.bas";
  != %basis
 GTOName ="system.bas";
 GTOAuxName ="system.aux";
 ]
 *** 1 failure is detected in the test module "orca_test"
```

junghans:
@JensWehner please review! Do we need tests?

felipeZ:
> I have two major concerns with this PR still
> 
> a) I do not get why we update the defaults with user options?
I think that updating the defaults with the user input is the normal behaviour that I would expect of a general simulation package. As an author of a simulation package, you don't want all your user to enter any single variable for configuring all the details of your simulation neither force them to make a copy of the whole options file and manually update the sections that the user are interested in. Instead, most simulation packages allow the user to provide minimal input that is subsequently merged with the whole set of configuration variables.

> b) I am not sure, if it is a good idea to default everything.

Put yourself on the user's shoes, you would like to try this amazing  XTP package to do a DFT-GWBSE simulation, if instead of having a very simple input interface (that can be modified for expert user) you force your users to enter every single detail of the simulation, the potential users would run away ;)
Also, we want to interface with Python using ASE. The philosophy of those wrappers is "plug and play", meaning that a minimal configuration is required to make a given package to run which is very useful when doing scripting.

JensWehner:
Okay I agree with a). 

Concerning b) I agree that most things can be defaulted, but inputfilenames should not be defaulted in my opinion.

felipeZ:
> Okay I agree with a).

How do you envision the input/defaults handling?

> Concerning b) I agree that most things can be defaulted, but inputfilenames should not be defaulted in my opinion.

How shall we then handle the input filenames?

JensWehner:
I thought we do it the same way as before. I agree that 90% of all options have a sensible default, but I do not think that the coordinates file, or the output file have that.

JensWehner:
Apart from eanalyze and ianalzye I cannot see how you would default all options of a calculator.

junghans:
@JensWehner please review.

PR #417, opened 2020-04-03
allow optional gwbse and dftpackage

felipeZ:
## Changes
Allow users to use the `dftpackage` and `gwbse_options` defaults by omitting those sections.


felipeZ:
I just realized what is the difference between `options` and `_options` variables

PR #421, opened 2020-04-07
Make options file optional for xtp_run

felipeZ:
See: #319 
## changed
The `options` file is now optional for the following `xtp_run` calculators:
   * [x] eanalyze
   * [x] einternal
   * [x] ianalyze
   * [x] kmclifetime
   * [x] kmcmultiple
   * [x] mapchecker
   * [x] neighborlist
   * [x] vaverage

### DIFF
--- a/share/xtp/xml/kmclifetime.xml
+++ b/share/xtp/xml/kmclifetime.xml
@@ -2,16 +2,16 @@
     <kmclifetime help="Kinetic Monte Carlo simulations of singlets with decay" label="calc:kmclifetime" section="sec:kmc">
         <numberofinsertions>4000</numberofinsertions>
         <seed help="Integer to initialise the random number generator" unit="integer" default=>23</seed>
-        <numberofcarriers help="Number of electrons/holes in the simulation box" unit="integer">1</numberofcarriers>
-        <injectionpattern help="Name pattern that specifies on which sites injection is possible. Use the wildcard '*' to inject on any site.">*</injectionpattern>
-        <lifetimefile help="File from which lifetimes are read in.">lifetimes.xml</lifetimefile>
-        <temperature help="Temperature in Kelvin.">300</temperature>
+        <numberofcarriers help="Number of electrons/holes in the simulation box" unit="integer" default=1>1</numberofcarriers>
+        <injectionpattern help="Name pattern that specifies on which sites injection is possible. Use the wildcard '*' to inject on any site." default="*">*</injectionpattern>
+        <lifetimefile help="File from which lifetimes are read in." default="lifetimes.xml">lifetimes.xml</lifetimefile>
+        <temperature help="Temperature in Kelvin." unit="kelvin" default=300>300</temperature>
         <trajectoryfile help="Name of the trajectory file" unit="">run1.csv</trajectoryfile>
         <carrierenergy help="Run a single carrier at the same time and records its energy">
-	        <run help="Switch on/off">0</run>
-	        <outputfile help="File to write energies to">energies.csv</outputfile>
-	        <alpha help="Smoothing energy profile">0.05</alpha>
-	        <outputsteps help="Write to file every x steps">10</outputsteps>
+	        <run help="Switch on/off" default="0">0</run>
+	        <outputfile help="File to write energies to" default="energies.csv">energies.csv</outputfile>
+	        <alpha help="Smoothing energy profile" default="0.05">0.05</alpha>
+	        <outputsteps help="Write to file every x steps" default="10">10</outputsteps>
         </carrierenergy>
     </kmclifetime>
 </options>
--- a/share/xtp/xml/kmclifetime.xml
+++ b/share/xtp/xml/kmclifetime.xml
@@ -1,7 +1,7 @@
 <options>
     <kmclifetime help="Kinetic Monte Carlo simulations of singlets with decay" label="calc:kmclifetime" section="sec:kmc">
         <numberofinsertions>4000</numberofinsertions>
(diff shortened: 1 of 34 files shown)


src/libxtp/calculators/ianalyze.cc
```diff
@@ -22,48 +22,42 @@
...
-    }
-  } else {
-    _states.push_back(QMStateType(QMStateType::Electron));
-    _states.push_back(QMStateType(QMStateType::Hole));
+void IAnalyze::Initialize(tools::Property &user_options) {
+
+  // get pre-defined default options from VOTCASHARE/xtp/xml/eanalyze.xml
+  LoadDefaults("xtp");
+  // update options with user specified input
+  UpdateWithUserOptions(user_options);
```
JensWehner, line 33: this should be at a lower level, because it is not special for ianalyze

src/libxtp/tools/dftgwbse.cc
```diff
@@ -24,8 +24,6 @@
 #include <votca/xtp/segment.h>
 #include <votca/xtp/staticregion.h>
 
-using namespace std;
```
JensWehner, line 27: nice

src/libxtp/calculators/ianalyze.cc
felipeZ: What do you mean by lower level :) ? like in a parent class method?

JensWehner: yes exactly. See below.

share/xtp/xml/kmclifetime.xml
```diff
@@ -1,17 +1,17 @@
 <options>
     <kmclifetime help="Kinetic Monte Carlo simulations of singlets with decay" label="calc:kmclifetime" section="sec:kmc">
         <numberofinsertions>4000</numberofinsertions>
-        <seed help="Integer to initialise the random number generator" unit="integer">23</seed>
-        <numberofcarriers help="Number of electrons/holes in the simulation box" unit="integer">1</numberofcarriers>
-        <injectionpattern help="Name pattern that specifies on which sites injection is possible. Use the wildcard '*' to inject on any site.">*</injectionpattern>
-        <lifetimefile help="File from which lifetimes are read in.">lifetimes.xml</lifetimefile>
-        <temperature help="Temperature in Kelvin.">300</temperature>
+        <seed help="Integer to initialise the random number generator" unit="integer" default=>23</seed>
+        <numberofcarriers help="Number of electrons/holes in the simulation box" unit="integer" default=1>1</numberofcarriers>
```
baumeier, line 5: default="1" 
I think the " are required.

share/xtp/xml/kmclifetime.xml
```diff
@@ -1,17 +1,17 @@
...
+        <injectionpattern help="Name pattern that specifies on which sites injection is possible. Use the wildcard '*' to inject on any site." default="*">*</injectionpattern>
+        <lifetimefile help="File from which lifetimes are read in." default="lifetimes.xml">lifetimes.xml</lifetimefile>
+        <temperature help="Temperature in Kelvin." unit="kelvin" default=300>300</temperature>
```
baumeier, line 8: same?

share/xtp/xml/neighborlist.xml
```diff
@@ -1,12 +1,9 @@
 <options>
-
-
 <neighborlist help="Determines neighbouring pairs for exciton transport" section="sec:neighborlist">
         <segments help="cutoff distances for certain segments identified by name">
-                <type help="segment name1 segment_name2">DCV5T_ME3 DCV5T_ME3</type>
-                <cutoff unit="nm">3</cutoff>
+                <type help="segment name1 segment_name2" default="DCV5T_ME3 DCV5T_ME3">DCV5T_ME3 DCV5T_ME3</type>
```
baumeier, line 4: Having the whole `segments` option here makes no sense, because there are now really system specific names given. We should make this part 100% optional

share/xtp/xml/neighborlist.xml
```diff
@@ -1,12 +1,9 @@
...
+                <cutoff unit="nm" default="3">3</cutoff>
         </segments>
-        <constant  help="cutoff for all other types of pairs" unit="nm">1.5</constant>
-        <exciton_cutoff help="cutoff beyond which pairs are only treated as singlet transition charges" unit="nm">0.5</exciton_cutoff>
+        <constant  help="cutoff for all other types of pairs" unit="nm" default="1.5">1.5</constant>
```
baumeier, line 7: This is the only entry that should be default.

src/tools/xtp_run.cc
```diff
@@ -107,18 +103,18 @@ bool XtpRun::EvaluateOptions() {
   for (const auto& calc : xtp::Calculators().getObjects()) {
 
     if (calc_string[0].compare(calc.first) == 0) {
-      cout << " This is a XTP app" << endl;
+      std::cout << " This is a XTP app\n";
```
baumeier, line 106: I think this can go. It is a remnant of when we were wrapping CTP calculators too

src/tools/xtp_tools.cc
```diff
@@ -124,17 +123,17 @@ bool XtpTools::EvaluateOptions() {
   bool found_calc = false;
   for (const auto& tool : xtp::QMTools().getObjects()) {
     if (calc_string[0].compare(tool.first) == 0) {
-      cout << " This is a XTP app" << endl;
+      std::cout << " This is a XTP app\n";
```
baumeier, line 126: same as above

Issue #341, opened 2019-12-05
unit_test_eeinteractor: fails on OpenSUSE


junghans:
From testing on x86_64 on OpenSUSE Tumbleweed:
```
[ 1136s] 57/69 Test #57: unit_test_eeinteractor ..................***Failed    0.03 sec
[ 1136s] Running 9 test cases...
[ 1136s] ref
[ 1136s]        -0
[ 1136s]        -0
[ 1136s] -0.333333
[ 1136s] dipole
[ 1136s]  -1.7459e+76
[ 1136s] -1.53098e+50
[ 1136s] -1.38781e+51
[ 1136s] ref
[ 1136s]        0
[ 1136s]        0
[ 1136s] 0.333333
[ 1136s] dipole
[ 1136s]   1.7459e+76
[ 1136s] -1.53098e+50
[ 1136s] -1.38781e+51
[ 1136s] /home/abuild/rpmbuild/BUILD/xtp-1.6_rc1/src/tests/test_eeinteractor.cc(490): [1;31;49merror: in "eeinteractor_test/polar_case_monopole": difference{1.37168e+153} between epolar.E_indu_indu(){-3.0481681366741134e+152} and 2.0 / 9.0{0.22222222222222221} exceeds 1e-12%[0;39;49m
[ 1136s] /home/abuild/rpmbuild/BUILD/xtp-1.6_rc1/src/tests/test_eeinteractor.cc(491): [1;31;49merror: in "eeinteractor_test/polar_case_monopole": difference{1.79769e+308} between epolar.E_indu_stat(){0} and -2.0 / 3.0{-0.66666666666666663} exceeds 1e-12%[0;39;49m
[ 1136s] /home/abuild/rpmbuild/BUILD/xtp-1.6_rc1/src/tests/test_eeinteractor.cc(492): [1;31;49merror: in "eeinteractor_test/polar_case_monopole": difference{2.74335e+153} between einternal{3.0481681366741134e+152} and 1.0 / 9.0{0.1111111111111111} exceeds 1e-12%[0;39;49m
[ 1136s] 
[ 1136s] [1;31;49m*** 3 failures are detected in the test module "eeinteractor_test"
[ 1136s] [0;39;49m
[ 1136s] 
```

[Details here](https://github.com/votca/xtp/files/3928798/build_xtp_opensuse_log.txt)

JensWehner:
hmm... looks like no initialisation or so. Hmmm

JensWehner:
[...]

PR #428, opened 2020-04-15
cast to scalar and changed polar distance

JensWehner:
Fix https://github.com/votca/xtp/issues/341


junghans:
Let me give it a test drive on OpenSuse

JensWehner:
yes.... :D

Issue #408, opened 2020-03-17
MO and Log file defaults


JensWehner:
the gwbseengine needs the names of the log file and the file which has the orbital information. 

i.e. for orca system.log and system.gbw
and for xtpdft only system_dft.orb as the logfile. it would be nice if they become part of the default options for qmpackages and do not have to be querried via the gwbseengine options anymore.

felipeZ:
see #414

PR #429, opened 2020-04-15
Name xtp tools

### LINKED ISSUE #426: improve defaults
## Proposed changes
* Introduce a new command-line option `-n` for the `xtp_calculator` with the name of the job.
* Derive the `xtp_tools` calculators filenames from the `-n` name argument

## Fix
* Allow running the tools without an option file (Currently this feature is broken)
* Remove the default file names from the XML files

felipeZ:
fix #426 and #408


junghans:
```
67/81 Test #211: unit_test_orca ............................***Failed    0.03 sec
Running 5 test cases...
Warning: VOTCASHARE environment variable not defined
unknown location(0): fatal error: in "orca_test/input_generation_version_4_0_1": std::out_of_range: basic_string::substr: __pos (which is 18446744073709551615) > this->size() (which is 0)
/home/votca/votca/xtp/src/tests/test_orca.cc(8634): last checkpoint
*** 1 failure is detected in the test module "orca_test"
```

baumeier:
I for one am positive about this change. For `xtp_tools` a lot of the filename specifications (including the extensions) in the xml files are dead weight and not at all clear for the non-expert user, as we have seen too often even internally. Specifying a command line option instead will mean for instance having the same call of `xtp_tools` for running `dftgwbse`, plotting the spectrum, fitting partial charges, and making a cube file, and not having to edit four xml files.

share/xtp/xml/gencube.xml
```diff
@@ -1,7 +1,5 @@
 <options>
 <gencube help="Tool to generate cube files from .orb file" section="sec:gencube">
-        <output help="Output file" default="votca.cube">votca.cube</output>
-        <input help="Input file" default="votca.orb">votca.orb</input>
```
JensWehner, line 4: So I cannot specify the output manually even if I want to?

felipeZ: of course, you can change the name of the input/output files,  see: https://github.com/votca/xtp/blob/name_xtp_tools/src/libxtp/tools/gencube.cc#L39
the `-n name` option is there just to help the user to avoid having to fill all the names in the XML files manually.

PR #451, opened 2020-05-18
mv bsecoupling defaults to the xml files

felipeZ:
## Proposed changes
Remove the default values from the header file of `bsecoupling`  into the Calculator XML files (e.g. `iqm`)


junghans:
Maybe add a test.

felipeZ:
> Maybe add a test.

@junghans I am creating a Jupyter notebook with the tutorial, where we can test this functionality

PR #466, opened 2020-06-13
using reductions instead of handcrafted resolutions

JensWehner:
Openmp is now required in version 4.5

this is not a big problem for gcc and icc but clang is a bit stupid, so you have to have clang-10 to use it properly. This seems to be a bug in the clang package.

### DIFF
--- a/include/votca/xtp/eigen.h
+++ b/include/votca/xtp/eigen.h
@@ -45,6 +45,13 @@ class Mat_p_Energy {
   Mat_p_Energy(double e, Eigen::MatrixXd&& mat)
       : _energy(e), _matrix(std::move(mat)){};
 
+  Mat_p_Energy operator+(const Mat_p_Energy& other) const {
+    Mat_p_Energy result = *this;
+    result._energy += other._energy;
+    result._matrix += other._matrix;
+    return result;
+  }
+
   Index rows() const { return _matrix.rows(); }
   Index cols() const { return _matrix.cols(); }
   Eigen::MatrixXd& matrix() { return _matrix; }
@@ -56,6 +63,21 @@ class Mat_p_Energy {
   double _energy;
   Eigen::MatrixXd _matrix;
 };
+
+#pragma omp declare reduction (+:Mat_p_Energy: omp_out=omp_out+omp_in)\
+     initializer(omp_priv=Mat_p_Energy(omp_orig.rows(),omp_orig.cols()))
+
+#pragma omp declare reduction (+: Eigen::VectorXd: omp_out=omp_out+omp_in)\
+     initializer(omp_priv=Eigen::VectorXd::Zero(omp_orig.size()))
+
+#pragma omp declare reduction (+: Eigen::MatrixXd: omp_out=omp_out+omp_in)\
+     initializer(omp_priv=Eigen::MatrixXd::Zero(omp_orig.rows(),omp_orig.cols()))
+
+#pragma omp declare reduction (+: Eigen::Matrix3d: omp_out=omp_out+omp_in)\
+     initializer(omp_priv=Eigen::Matrix3d::Zero())
+
+#pragma omp declare reduction (+: Eigen::Vector3d: omp_out=omp_out+omp_in)\
+     initializer(omp_priv=Eigen::Vector3d::Zero())
 
 namespace OPENMP {
 inline Index getMaxThreads() {
--- a/include/votca/xtp/matrixfreeoperator.h
+++ b/include/votca/xtp/matrixfreeoperator.h
@@ -166,27 +166,25 @@ struct generic_product_impl<votca::xtp::MatrixFreeOperator, Mtype, DenseShape,
       Index blocks = op.size() / blocksize;
       // this uses the fact that all our matrices are symmetric, i.e we can
       // reuse half the blocks
-      std::vector<Eigen::MatrixXd> thread_wiseresult(
-          votca::xtp::OPENMP::getMaxThreads(),
-          Eigen::MatrixXd::Zero(dst.rows(), dst.cols()));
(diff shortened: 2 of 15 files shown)


JensWehner:
So clang is problematic. Hmmm...

src/libxtp/CMakeLists.txt
```diff
@@ -10,7 +10,11 @@ list(REMOVE_ITEM VOTCA_SOURCES ${NOT_VOTCA_SOURCES})
 add_library(votca_xtp  ${VOTCA_SOURCES})
 set_target_properties(votca_xtp PROPERTIES SOVERSION ${SOVERSION})
 add_dependencies(votca_xtp gitversion-xtp)
-target_link_libraries(votca_xtp PUBLIC VOTCA::votca_csg VOTCA::votca_tools Boost::boost Eigen3::Eigen ${HDF5_LIBRARIES} PRIVATE LIBXC::LIBXC Boost::program_options Boost::filesystem Boost::system Boost::timer )
+target_link_libraries(votca_xtp PUBLIC VOTCA::votca_csg VOTCA::votca_tools Boost::boost  Eigen3::Eigen ${HDF5_LIBRARIES} PRIVATE LIBXC::LIBXC Boost::program_options Boost::filesystem Boost::system Boost::timer )
```
junghans, line 13: Why that extra space? ;-)

include/votca/xtp/eigen.h
```diff
@@ -45,6 +45,13 @@ class Mat_p_Energy {
   Mat_p_Energy(double e, Eigen::MatrixXd&& mat)
       : _energy(e), _matrix(std::move(mat)){};
 
+  Mat_p_Energy operator+(const Mat_p_Energy& other) const {
```
felipeZ, line 48: Really nice!

src/tests/test_threecenter_dft.cc
```diff
@@ -51,8 +51,6 @@ Eigen::MatrixXd ReadMatrixFromString(const std::string& matrix) {
 
 BOOST_AUTO_TEST_CASE(small_basis) {
 
-  OPENMP::setMaxThreads(1);
```
felipeZ, line 54: Why is it now possible to use multiple threads here?

src/libxtp/dftengine/ERIs.cc
```diff
@@ -152,16 +136,14 @@ Mat_p_Energy ERIs::CalculateERIs_4c_small_molecule(
...
-  std::vector<Eigen::MatrixXd> EXX_thread = std::vector<Eigen::MatrixXd>(
-      OPENMP::getMaxThreads(), Eigen::MatrixXd::Zero(DMAT.rows(), DMAT.cols()));
+  Eigen::MatrixXd EXX = Eigen::MatrixXd::Zero(DMAT.rows(), DMAT.cols());
 
   const Eigen::VectorXd& fourc_vector = _fourcenter.get_4c_vector();
 
   Index dftBasisSize = DMAT.rows();
   Index vectorSize = (dftBasisSize * (dftBasisSize + 1)) / 2;
-#pragma omp parallel for
+#pragma omp parallel for reduction(+ : EXX)
```
felipeZ, line 145: The readability has improved significantly!

include/votca/xtp/matrixfreeoperator.h
```diff
@@ -166,27 +166,25 @@ struct generic_product_impl<votca::xtp::MatrixFreeOperator, Mtype, DenseShape,
       Index blocks = op.size() / blocksize;
       // this uses the fact that all our matrices are symmetric, i.e we can
       // reuse half the blocks
-      std::vector<Eigen::MatrixXd> thread_wiseresult(
-          votca::xtp::OPENMP::getMaxThreads(),
-          Eigen::MatrixXd::Zero(dst.rows(), dst.cols()));
-#pragma omp parallel for schedule(guided)
+      Eigen::MatrixXd result = Eigen::MatrixXd::Zero(dst.rows(), dst.cols());
+
+#pragma omp declare reduction (+: Eigen::MatrixXd: omp_out=omp_out+omp_in)\
```
felipeZ, line 171: I don't fully understand this snippet, could you please tell me how it works?

felipeZ: `omp_out=omp_out+omp_in` is the communicative operation to perform and `omp_priv` is the "zero" element, right?

felipeZ: if the `reduction` for `MatrixXd` has been declared in `eigen.h` why don't you use it here?

src/tests/test_threecenter_dft.cc
JensWehner: I am not sure why it was disabled before.

include/votca/xtp/matrixfreeoperator.h
JensWehner: this is in the eigen::internal namespace, so I had to do it again.

src/libxtp/CMakeLists.txt
JensWehner: ups

src/libxtp/CMakeLists.txt
```diff
@@ -10,7 +10,11 @@ list(REMOVE_ITEM VOTCA_SOURCES ${NOT_VOTCA_SOURCES})
...
+
+if(TARGET OpenMP::OpenMP_CXX)
+  target_link_libraries(votca_xtp PUBLIC OpenMP::OpenMP_CXX)
```
junghans, line 16: We are planning on any consumers for xtp if so we might need to add this to CMake exported target as well.

JensWehner: I do not think there is a library which uses xtp but yeah.

PR #467, opened 2020-06-15
GitHub actions

junghans:


.github/workflows/continuous-integration-workflow.yml
```diff
@@ -0,0 +1,84 @@
...
+            no_regression_testing: true
+    runs-on: ubuntu-latest
+    container: votca/buildenv:${{ matrix.distro }}
+    steps:
+      - uses: actions/checkout@v2.2.0
+        with:
+          submodules: true
+      - name: Run Votca Setup
+        id: setup
+        uses: votca/actions/setup@master
```
felipeZ, line 42: @junghans this [votca/actions](https://github.com/votca/actions/blob/master/setup/action.yml) just setup all the CMake variables *inside* the ``votca/buildenv:${{ matrix.distro }}``, am I right?

junghans: Correct! It also checks out the code of the other modules.

junghans: Actually, the action is always done inside `votca/buildenv:latest`, which sometimes leads to pull of two different containers.

PR #470, opened 2020-06-16
Create a map between Orca and LibXC names

### LINKED ISSUE #461: libxc and Orca functional names
## Description
For running a DFT-GWBSE calculations the user needs to provide a **XC** functional but the problem is that [Orca uses its own functional names](https://sites.google.com/site/orcainputlibrary/dft) and for the GWBSE calculations **XTP** follows the **libXC**  name convention. So, the current configuration breaks if we try to use Orca and then perform a GWBSE calculation

## Solutions
* **Use libXC name convention**: Create and maintain a map from XC  to Orca functional names.
* **Use Orca name convention**: Since Orca 4.2 It is printed in the Orca output the name in the libXC convention, so we can parse and use it.

Since we are also planning to integrate [PySCF](https://sunqm.github.io/pyscf/) and it uses **libXC** then option 1 since more suitable.

@baumeier @JensWehner  any thoughts about this?

felipeZ:
## Changes
Fix #461 by adding the ``share/packages/orca_functional_names.xml`` containing the naming correspondence between orca and libxc functionals.


share/packages/orca_functional_names.xml
```diff
@@ -0,0 +1,7 @@
+<?xml version="1.0"?>
+<functionals>
+  <XC_HYB_GGA_XC_B1LYP number="416">B1LYP</XC_HYB_GGA_XC_B1LYP>
+  <XC_HYB_GGA_XC_B3LYP number="402">B3LYP</XC_HYB_GGA_XC_B3LYP>
+  <XC_HYB_GGA_XC_PBEH number="406">PBE0</XC_HYB_GGA_XC_PBEH>
+  <XC_GGA_C_PBE number="130">PBE</XC_GGA_C_PBE>
```
JensWehner, line 6: this line is problemati because the PBE functional is XC_GGA_C_PBE + XC_GGA_X_PBE,

C and X indicate the correlation and exchange part separately.

src/libxtp/qmpackages/orca.cc
```diff
@@ -825,12 +825,38 @@ std::string Orca::WriteMethod() const {
...
          // additional properties provided by the user
          << _settings.get("orca.method") << "\n";
   return stream.str();
 }
 
+std::string Orca::GetOrcaFunctionalName() const {
+
+  char* votca_share = getenv("VOTCASHARE");
+  if (votca_share == nullptr) {
+    return _settings.get("functional");
```
JensWehner, line 839: I do not understand the logic here.

src/libxtp/qmpackages/orca.cc
```diff
@@ -825,12 +825,38 @@ std::string Orca::WriteMethod() const {
...
+
+    const tools::Property& functional_names =
+        all_functionals.get("functionals");
+
+    std::string input_name = _settings.get("functional");
+
+    if (functional_names.exists(input_name)) {
+      return functional_names.get(input_name).as<std::string>();
+    } else {
+      return input_name;
```
JensWehner, line 856: should we not throw an error here?

src/libxtp/qmpackages/orca.cc
felipeZ: The `VOTCASHARE` environmental variable is not defined for testing, so we need to take into account that  corner case

src/libxtp/qmpackages/orca.cc
felipeZ: I am not sure if that is the best solution. We will throw an error based on the assumption that *Orca* doesn't know the functional that the user has passed based on an small subset of functional that we have. But I suppose that for most of the users an error in this point is more meaningful that in the Orca runtime.
I am going to change it.

share/packages/orca_functional_names.xml
felipeZ: mmmm, so when you are using PBE do you call functional as:
 ```
XC_GGA_C_PBE+XC_GGA_X_PBE
```
Or how do you call it?

File: src/libxtp/qmpackages/orca.cc (hunk 2 above)
Comment (JensWehner) [reply src/libxtp/qmpackages/orca.cc:839]: yeah environment variables and testing are a shit combination. Maybe we have to think about a general solution for that. Any ideas?

File: share/packages/orca_functional_names.xml (hunk 1 above)
Comment (JensWehner) [reply share/packages/orca_functional_names.xml:6]: XC_GGA_C_PBE XC_GGA_X_PBE

Comment (JensWehner) [reply share/packages/orca_functional_names.xml:6]: the order does not matter

File: src/libxtp/qmpackages/orca.cc [hunk 4]
```diff
@@ -825,12 +825,51 @@ std::string Orca::WriteMethod() const {
...
+    auto xml_file = std::string(getenv("VOTCASHARE")) +
+                    std::string("/xtp/packages/orca_functional_names.xml");
+
+    all_functionals.LoadFromXML(xml_file);
+
+    const tools::Property& functional_names =
+        all_functionals.get("functionals");
+
+    std::string input_name = _settings.get("functional");
+    // Some functionals have a composed named separated by a space
```
felipeZ, line 852: @JensWehner  do you agree with this?

JensWehner: what happens if votcashare is not set?

felipeZ: Assuming that the user has command line tools in her/his path and the calculation doesn't fail because ti cannot find `xtp_tools`, `xtp_run`, etc.then that function will generate a call to Orca that will subsequently fails because it doesn't know the functional

felipeZ:
@JensWehner  @baumeier  Could you please review this PR?

JensWehner:
Should we maybe introduce a new folder which stores, qmdata default?

felipeZ:
> Should we maybe introduce a new folder which stores, qmdata default?

@JensWehner  I have done so at 879dc1a. I also added some README files and deleted the used `man` folder

felipeZ:
@JensWehner Testing the LibXC/Orca functional name conversion is difficult because of the `VOTCASHARE` environmental variable and the fact that all the new methods are private.

PR #476, opened 2020-06-18
Xtpshare

### LINKED ISSUE #475: Remove xtp/share/package folder
## The problem
Currently there are two folder where XTP stores configuration files:
```
share/packages
share/xtp/xml
```
The content of the first folder is installed at `${VOTCASHARE}/votca/xtp/packages` and the second to 
`${VOTCASHARE}/votca/xtp/xml`

Both the actual source path and the installation path are used in different parts for configuration, installation, running the CI and running the actual calculation by the user.
These paths are the cause of many misconfigurations and problems calling the runners.

For example, the `votca_xtp` target properties are set like:
```cmake
get_filename_component(_PARENT_PATH "${CMAKE_CURRENT_BINARY_DIR}" DIRECTORY)

add_library(votca_xtp  ${VOTCA_SOURCES})
set_target_properties(votca_xtp PROPERTIES SOVERSION ${SOVERSION})
set_target_properties(votca_xtp PROPERTIES VOTCA_SHARE ${CMAKE_SOURCE_DIR}/xtp/share)
set_target_properties(votca_xtp PROPERTIES XTP_PATH ${_PARENT_PATH}/tools)
```
The `VOTCA_SHARE` variable is pointing correctly to the calculators configuration but erroneously to the package path.

## The solution
Move the content of `share/packages` into `share/xtp/xml`.

## TODO
* [ ] move the xml files from one folder to the other
* [ ] fix all the broken path invoking the calculator
* [ ] fix the documentation generation
* [ ] fix the target used by script calling the Jupyter notebooks in the CI

felipeZ:
fix #475


src/libxtp/CMakeLists.txt
```diff
@@ -9,9 +9,10 @@ list(REMOVE_ITEM VOTCA_SOURCES ${NOT_VOTCA_SOURCES})
 
 get_filename_component(_PARENT_PATH "${CMAKE_CURRENT_BINARY_DIR}" DIRECTORY)
 
+message(STATUS: "PARENT_PATH: ${_PARENT_PATH}")
 add_library(votca_xtp  ${VOTCA_SOURCES})
 set_target_properties(votca_xtp PROPERTIES SOVERSION ${SOVERSION})
-set_target_properties(votca_xtp PROPERTIES VOTCA_SHARE ${CMAKE_SOURCE_DIR}/xtp/share)
+set_target_properties(votca_xtp PROPERTIES VOTCA_SHARE ${_PARENT_PATH}/../share)
```
junghans, line 15: I would use `${PROJECT_SOURCE_DIR}/share` here.

felipeZ: The problem with using `${PROJECT_SOURCE_DIR}/share` is doesn't contain the basisset. So,  I have created a temporal folder at [share](https://github.com/votca/xtp/blob/xtpshare/share/CMakeLists.txt) where I collect all the the data that need to be on `VOTCASHARE`.

junghans: Where do they get created? `${PROJECT_BINARY_DIR}/share`?

felipeZ: They XML files already exist in different locations of the `votca/xtp/share` folder that **do not correspond one to one with the final installation folder**. During the installation they are all put under share/xtp. So, I was wondering if I can use the installation `share` path because they are in the correct subfolders.

junghans: I see, you need the xml files in`${VOTCASHARE}/xtp/xml`. How about moving `share/xml` to `share/xtp/xml`, then you can set `VOTCA_SHARE` above to `${PROJECT_SOURCE_DIR}/share`.

felipeZ: moving the `share/basis_set` to `share/xtp/basis_set` made the trick, the jupyter notebooks now use the `${PROJECT_SOURCE_DIR}/share` folder as `VOTCASHARE` and run properly (at least locally)

felipeZ:
@JensWehner can I merge this?

PR #478, opened 2020-06-19
Storing orca gbw files in eqm for merged guess 

### LINKED ISSUE #472: gbw files are not copied into the right folder.
In iqm the gbw files generated with eqm are not copied in the right folders. I spot the problem in the following section
https://github.com/votca/xtp/blob/0c5dd843715f3b0d48faf04d7fd43e07f21c540c/src/libxtp/jobcalculators/iqm.cc#L329

baumeier:
fixes #472


baumeier:
If you mean that this puts the files where `iqm` expects them, then it cooperates well.

Issue #463, opened 2020-06-13
MKL backend not used in Eigen


baumeier:
It appears as if the step of `Diagonalizing the two-particle Hamiltonian` in the exact FAA approach for the self-energy is currently only running in serial. Seen both with `gcc` and native `Eigen` backend, as well as with `icc` and `mkl`.

JensWehner:
the rest is parallel? It sounds a bit like a problem with the linking to MKL. With native Eigen matrix diagonalisation is not parallelized.

baumeier:
Rest is running in parallel

JensWehner:
hmmm.. what kind of threading is mkl using?

baumeier:
`cat CMakeCache.txt | grep MKL`

```
//The thread layer to choose for MKL
MKL_THREAD_LAYER:STRING=Intel OpenMP
MKL_ThreadLayer_LINK_LIBRARY:FILEPATH=/opt/intel/mkl/lib/intel64/libmkl_intel_thread.so
MKL_ThreadLayer_STATIC_LINK_LIBRARY:FILEPATH=/opt/intel/mkl/lib/intel64/libmkl_intel_thread.a
MKL_ThreadingLibrary_LINK_LIBRARY:FILEPATH=/usr/lib/x86_64-linux-gnu/libiomp5.so
MKL_ThreadingLibrary_STATIC_LINK_LIBRARY:FILEPATH=/opt/intel/lib/intel64/libiomp5.a
```

Is it a problem that it uses the system `libiomp5.so`?

JensWehner:
yes that could be, I do not know.

JensWehner:
where is it coming from that is not the openmp library?

baumeier:
I manually set

`MKL_ThreadingLibrary_LINK_LIBRARY:FILEPATH=/opt/intel/mkl/lib/intel64/libiomp5.so`

and checked with `ldd`

```
	libvotca_xtp.so.7 => /home/bbaumeie/work/votca_intel/lib/libvotca_xtp.so.7 (0x00007f730f3f1000)
	libvotca_csg.so.7 => /home/bbaumeie/work/votca_intel/lib/libvotca_csg.so.7 (0x00007f730f24b000)
	libvotca_tools.so.7 => /home/bbaumeie/work/votca_intel/lib/libvotca_tools.so.7 (0x00007f730f108000)
[...]

PR #479, opened 2020-06-19
Mkl fix

JensWehner:
Should fix https://github.com/votca/xtp/issues/463

This should fix the issue, I have not found the root cause yet, the log output should be more reliable now.


JensWehner:
@baumeier can you test this?

baumeier:
@JensWehner works

PR #480, opened 2020-06-20
4th row elements support

baumeier:
Defines integration grid for selected heavy elements and the associated ECPs, basis sets.


share/ecps/def2-ecp.xml
```diff
@@ -0,0 +1,133 @@
+<pseudopotential name="def2-ecp">
```
JensWehner, line 1: could you put this through an xml formatter?

baumeier: Done.

PR #489, opened 2020-06-24
cmake: export xtp target

junghans:
For https://github.com/votca/xtp-tutorials/pull/19


junghans:
@votca-bot changelog: add cmake exported target

PR #491, opened 2020-06-29
Fix failing tests on Fedora 32

junghans:


junghans:
This PR is mainly to reproduce the issue from https://github.com/votca/votca/pull/350, but it seems it doesn't get triggered here.....

junghans:
See #493

PR #496, opened 2020-07-01
CI: add mkl builds

junghans:


junghans:
@votca-bot changelog: add mkl builds to CI

PR #499, opened 2020-07-04
cmake: enable CXX only

junghans:


junghans:
@votca-bot changelog: enable CXX only in CMake

PR #500, opened 2020-07-06
CI: do not build for master

junghans:


JensWehner:
you do not want to build stuff twice anymore, because nobody should push to master anyway?

junghans:
correct

junghans:
Plus, the update will be forwarded to votca/votca automatically  anyhow and tested there.

Hmm, that effectively makes the badge in the README useless....

JensWehner:
true.

junghans:
Regarding the badge we could use [CI](https://github.com/votca/xtp/actions?query=workflow%3ACI) instead, which would just show all CI builds.

junghans:
Problem is that the badge could turn red, when the last CI run on a PR fails, but it should be mostly green.

Any opinion?

JensWehner:
can we not use the votca/votca badge?

junghans:
Sure

JensWehner:
I would think that is the best option then.

junghans:
Agreed, the CI in `votca/votca` is the only thing that really matters anyhow.

JoshuaSBrown:
Fine by me.

junghans:
Hmm for coverage reporting we will do at least run the coverage builds on master. Problem is that one cannot exclude a part of the matrix depending wether it is a triggered by a PR or a push. 

3 options to re-enable the coverage builds on master 1) replicate half of the CI file in new file (would be `diff` friendly), add a copy to the same file (has a lot of duplication as well) as new jobs. 3) conditional are possible for each step of a job, so disable each step for the unwanted jobs on master (less duplication, but a lot of `if:` and the job will still get triggered).

I am leaning towards 1) @JensWehner any opinion?

JensWehner:
I think 1)is the best choice

Issue #498, opened 2020-07-04
RPA correlation energy


baumeier:
Calculate correlation energy in the RPA if FAA is used. See https://aip.scitation.org/doi/10.1063/1.3442749 Eq.9.

This could be pretty straightforward?

PR #501, opened 2020-07-07
RPA correlation energy in FAA

baumeier:
implements #498 
@JensWehner please check if the part regarding the contribution of the diagonal elements of the RPA Hamiltonian is correct. Numbers agree with reference calculations.


src/libxtp/dftengine/dftengine.cc
```diff
@@ -297,6 +297,25 @@ bool DFTEngine::Evaluate(Orbitals& orb) {
...
+      Mat_p_Energy EXXs = CalcEXXs(MOs.eigenvectors(), Dmat);
+      exx = -1.0 / 4.0 * EXXs.energy();
+      XTP_LOG(Log::error, *_pLog) << TimeStamp() << std::setprecision(12)
+                                  << " EXX energy " << exx << " Ha" << flush;
+
+      XTP_LOG(Log::error, *_pLog)
+          << TimeStamp() << std::setprecision(12) << " Final EXX Total energy "
+          << totenergy - e_vxc.energy() + (1.0 - _ScaHFX) * exx << " Ha"
+          << flush;
+
```
JensWehner, line 318: how does this relate to the change and what is it supposed to do?

src/libxtp/gwbse/rpa.cc
```diff
@@ -113,11 +115,17 @@ RPA::rpa_eigensolution RPA::Diagonalize_H2p() const {
...
+  RPA_correlation_energy += 0.5 * sol.omega.sum();
+
   XTP_LOG(Log::info, _log) << TimeStamp()
                            << " Lowest neutral excitation energy (eV): "
                            << tools::conv::hrt2ev * sol.omega.minCoeff()
                            << std::flush;
 
+  XTP_LOG(Log::error, _log) << TimeStamp()
+                           << " RPA correlation energy (Hartree): "
+                           << RPA_correlation_energy << std::flush;
```
JensWehner, line 127: could you add a comment  where this is coming from?

src/libxtp/dftengine/dftengine.cc
baumeier: This will give the user the opportunity to calculate the total energy of the system in the RPA approximation. For that you need to subtract the Exc contribution from the total DFT energy, add HF-like Ex unscaled. This yields the so called EXX energy (eXact eXchange) to which the RPA correlation energy needs to be added. See Eq.1 in https://aip.scitation.org/doi/10.1063/1.3442749 (they call EXX EHF)

E^RPA = E^XX + E^RPA_correlation

src/libxtp/gwbse/rpa.cc
baumeier: How the energy is calculated? Eq.9 same paper -1/2 Tr(HRPA^1/2 - A) as a code comment or output info ?

src/libxtp/dftengine/dftengine.cc
JensWehner: so the RPA stuff only works with xtpdft? How do you want to add this to the RPA? Or is the user supposed to add the contributions up by hand to get E^RPA

src/libxtp/gwbse/rpa.cc
JensWehner: comment in the code with the paper is good.

JensWehner: would be also nice to have some sort of test for this.

src/libxtp/dftengine/dftengine.cc
JensWehner: I do not really understand the workflow here.

baumeier: Yes, unless we want to make the EXX energy part of orbitals. I suppose one could get something out of orca's logfile but it is nasty if for example the DFT calculation was done without hybrid.

src/libxtp/gwbse/rpa.cc
baumeier: Not sure how since this is not stored/reported anywhere outside this method

baumeier: Unless we add an energy variable to `struct rpa_eigensolution`

src/libxtp/dftengine/dftengine.cc
JensWehner: hmm.. would it be possible to simply calculate this in the gw part instead of the DFT?

baumeier: Maybe. Vxc for sure, not immediately clear what about the exchange energy, since I think that is part of Sigma, and not returning the total energy.

JensWehner: you can add another function to sigma.

src/libxtp/gwbse/rpa.cc
baumeier: test added

src/libxtp/dftengine/dftengine.cc
baumeier: That's above my paygrade

JensWehner: ? I think you can calculate Exx from the Sigma_x function via:

<a href="https://www.codecogs.com/eqnedit.php?latex=E^{xx}=Tr(D*V^{xx})&space;=>&space;V^{xx}_{\alpha\beta}=\sum_{\gamma\delta}D_{\gamma\delta}(\alpha\gamma|\beta\delta)=>&space;D_{\alpha\beta}=\sum_i^{occ}&space;M_{\alpha&space;i}M_{i\beta}" target="_blank">&space;V^{xx}_{\alpha\beta}=\sum_{\gamma\delta}D_{\gamma\delta}(\alpha\gamma|\beta\delta)=>&space;D_{\alpha\beta}=\sum_i^{occ}&space;M_{\alpha&space;i}M_{i\beta}" title="E^{xx}=Tr(D*V^{xx}) => V^{xx}_{\alpha\beta}=\sum_{\gamma\delta}D_{\gamma\delta}(\alpha\gamma|\beta\delta)=> D_{\alpha\beta}=\sum_i^{occ} M_{\alpha i}M_{i\beta}" /></a>

if you put that all in you should get a nice function for Exx in terms of (ij|kl),i.e. the 4center integrals in MO basis, which are used in the GW calculation

baumeier: Yes, in principle. I practice I am not so sure what happens if the GW corrections are not requested for all occupied states, for instance, and the the 4c in MO basis are not the same in Sigma_x as needed for E_x.

baumeier: Although this maybe on RPAlevels level, but I have no time to dig through that now.

JensWehner: If you do not do a full range calculation you get the same problem for the correlation contribution as well.

baumeier: Not exactly the same. You may have an E_c calculated with a limited range set (which one might physically) but at least the E_x is the same as used in the hybrid DFT.

baumeier: I will merge this as is and put up another issue to extend the EXX total energy calculation to other codes by either having it in the qmpackage or in the Sigma.

PR #502, opened 2020-07-07
CI: turn coverage into own action

junghans:
https://github.com/votca/xtp/pull/500#issuecomment-654583279

### DIFF
--- a/.github/workflows/coverage.yml
+++ b/.github/workflows/coverage.yml
@@ -1,28 +1,21 @@
-name: CI
+name: Coverage
 on:
+  push:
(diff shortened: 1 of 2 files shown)


junghans:
Hmm, the problem with two workflows is that the coverage workflow will still keep running, if the CI workflow already failed.

junghans:
But writing them into one yaml as two jobs won't automatically cancel them either. We could make one job depend on the other using `depend:` just that would artificially serialize the execution. 

So let's leave it like that for now until we find a better solution.

Issue #492, opened 2020-06-30
Print all found quasiparticle energies and QP weights


baumeier:
The QP equations in the GW step may have several possible solutions. Currently the one with the largest spectral weight is chose as "THE" solution but it would be informative to also - maybe in verbose mode - also print the discarded solutions and their weights.

JensWehner:
How can I help?

baumeier:
You could either do it :) or let me know where this is exactly and if this already kept track of in some form or not

JensWehner:
I am going to have a look on wednesday.

PR #503, opened 2020-07-08
added qp root output

JensWehner:
implements https://github.com/votca/xtp/issues/492

due to the parallelization, it is not guaranteed that the output is ordered by level index, but the index->qp_weight,qp_energy is always correctly output.


JensWehner:
@votca-bot changelog: added verbose option for rootfinder

junghans:
```
      Start 226: unit_test_gw
112
54/82 Test #226: unit_test_gw ..............................***Failed    0.05 sec
```

PR #506, opened 2020-07-09
Properly write point charges in Orca

### LINKED ISSUE #504: Orca pointcharges are not generated properly
## Problem
During a QMMM calculation charges can be given to orca for reading using the filename `background.crg`.  But the method in charge of doing so 
 ```cc
void Orca::WriteChargeOption()
```
Fails to properly write that in the orca input.

## Solution 
Fix the `WriteChargeOption` method

felipeZ:
Fix #504 by properly *pointcharges* in the Orca input.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -48,6 +48,7 @@ Version 1.7-dev
 -  enable CXX only in CMake (#499)
 -  implement RPA correlation energy (#498)
 -  added verbose option for rootfinder (#503)
+-  fix orca pointcharges (#504)
 
 Version 1.6.1 (released 21.06.20)
 =================================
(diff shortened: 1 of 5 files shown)


felipeZ:
> I think we need a proper test of the orca input file. Otherwise good work.

The `Orca::WriteChargeOption()` method is private and it is only called inside the `QMPackage` method: `void AddRegion(const MMRegion& mmregion)` for the QMMM simulations.  So, for testing it, I would need to add a region and then generate the Orca input file or I could  make a `MockOrca` subclass derived from Orca and check that it is writing the `pointcharges` properly, but since Orca objects are generated with a Factory, deriving  a class from Orca without registering is not possible or at least not easy. Any suggestion?

JensWehner:
AHh that is true, hmm..... I do not know then.

felipeZ:
@JensWehner I cannot merge this PR because I don't have authorization. Could you please merge this branch?

Issue #508, opened 2020-07-15
Segment type based neighborlist


baumeier:
Segment type based neighborlist generation is not working. As the default calculator options define a `<constant>` cutoff, the bool `_useConstantCutoff` always evaluates to `true` in `neighborlist.cc` and the segment type part is never used.

felipeZ:
@baumeier I think that removing the `constant` property from the defaults should be enough. I need to add it on the tutorial

JensWehner:
There are some calculators which change behaviour if a tag is present or not, that collides with having default values.

felipeZ:
> There are some calculators which change behaviour if a tag is present or not, that collides with having default values.

I have removed the default, shall we do something else?

felipeZ:
solved by #509

PR #509, opened 2020-07-15
remove constant from neighborlist defaults

felipeZ:
## problem
see issue #508 

## proposed solution
Remove the `constant` keyword from the defaults of the `neighborlist` calculator.

### DIFF
--- a/share/xtp/xml/molpol.xml
+++ b/share/xtp/xml/molpol.xml
@@ -7,8 +7,8 @@
       <max_iter help="Maximum number of iterations" default="500"/>
       <exp_damp help="Thole sharpness parameter" default="0.39"/>
     </options_polar>
-    <mode help="Read the polar target from a file or a qmpackage logfile" choices="target,qmpackage"/>
-    <target help="should have this format: pxx pxy pxz pyy pyz pzz"/>
+    <mode help="Read the polar target from a file or a qmpackage logfile" choices="file,qmpackage"/>
+    <target_polarisability help="should have this format: pxx pxy pxz pyy pyz pzz"/>
     <qmpackage help="qmpackage name if mode is qmpackage" default="orca" choices="orca"/>
     <logfile help="logfile of DFTpackage to read Target polarisability from if mode is qmpackage" default="system.log"/>
     <tolerance help="convergence tolerance" default="1e-4" choices="float+"/>
--- a/share/xtp/xml/molpol.xml
+++ b/share/xtp/xml/molpol.xml
@@ -7,8 +7,10 @@
       <max_iter help="Maximum number of iterations" default="500"/>
       <exp_damp help="Thole sharpness parameter" default="0.39"/>
     </options_polar>
-    <qmpackage help="DFT package polarization calculation was run with" default="orca" choices="orca"/>
-    <logfile help="logfile of DFTpackage to read Target polarisability from" default="votca.orb"/>
+    <mode help="Read the polar target from a file or a qmpackage logfile" choices="target,qmpackage"/>
+    <target help="should have this format: pxx pxy pxz pyy pyz pzz"/>
+    <qmpackage help="qmpackage name if mode is qmpackage" default="orca" choices="orca"/>
+    <logfile help="logfile of DFTpackage to read Target polarisability from if mode is qmpackage" default="votca.orb"/>
     <tolerance help="convergence tolerance" default="1e-4" choices="float+"/>
     <iterations help="maximum number of iterations" default="100" choices="int+"/>
   </molpol>
--- a/share/xtp/xml/molpol.xml
+++ b/share/xtp/xml/molpol.xml
@@ -10,7 +10,7 @@
     <mode help="Read the polar target from a file or a qmpackage logfile" choices="target,qmpackage"/>
     <target help="should have this format: pxx pxy pxz pyy pyz pzz"/>
     <qmpackage help="qmpackage name if mode is qmpackage" default="orca" choices="orca"/>
-    <logfile help="logfile of DFTpackage to read Target polarisability from if mode is qmpackage" default="votca.orb"/>
+    <logfile help="logfile of DFTpackage to read Target polarisability from if mode is qmpackage" default="system.log"/>
     <tolerance help="convergence tolerance" default="1e-4" choices="float+"/>
     <iterations help="maximum number of iterations" default="100" choices="int+"/>
   </molpol>
--- a/share/xtp/xml/neighborlist.xml
+++ b/share/xtp/xml/neighborlist.xml
@@ -1,11 +1,11 @@
 <?xml version="1.0"?>
 <options>
   <neighborlist help="Determines neighbouring pairs for exciton transport" section="sec:neighborlist">
-    <type help="segments names if the segmentname method is used"/>
+    <segmentname help="segments names if the segmentname method is used"/>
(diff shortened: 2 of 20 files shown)


felipeZ:
> How much sense does it make to have a default optionsfile without defaults?

There are no defaults but there is some explanation what is the calculator about.
neighborlist


> Related question: How does this affect the auto-generated documentation? As I understand it, the documentation for the
> `neighborlist` calculator will now not list any options at all.

I am changing the QMMM tutorials by adding the `constant` section.

felipeZ:
> The option structure has to be redesigned

Could you please elaborate more about this?

JensWehner:
WIth the default values, an entry will always be added to the property,

so any calculator, which checks:

`if(opt.exists("optionA")` should not work anymore. 
instead there should probably be a

`<cutoff>constant</cutoff>` or `<cutoff>segmentname</cutoff>` which explicilty states which method should be used. 

So basicallly for all tools,calculators and qmcalculators any `opt.exists` stuff should be banned.

felipeZ:
Also solves #511 by introducing new flags and remove the `exists` calls.

felipeZ:
@JensWehner  do you have further thoughts about #508 and #511?

JensWehner:
what do you mean?

felipeZ:
> what do you mean?

I have removed the `exists` call from the calculators and introduced some new flags, do you agree with that?

share/xtp/xml/molpol.xml
```diff
@@ -7,8 +7,10 @@
       <max_iter help="Maximum number of iterations" default="500"/>
       <exp_damp help="Thole sharpness parameter" default="0.39"/>
     </options_polar>
-    <qmpackage help="DFT package polarization calculation was run with" default="orca" choices="orca"/>
-    <logfile help="logfile of DFTpackage to read Target polarisability from" default="votca.orb"/>
+    <mode help="Read the polar target from a file or a qmpackage logfile" choices="target,qmpackage"/>
+    <target help="should have this format: pxx pxy pxz pyy pyz pzz"/>
+    <qmpackage help="qmpackage name if mode is qmpackage" default="orca" choices="orca"/>
+    <logfile help="logfile of DFTpackage to read Target polarisability from if mode is qmpackage" default="votca.orb"/>
```
JensWehner, line 13: the defaults do not work here, the default should be `system.log` or so.

felipeZ: Thanks for let me know!

baumeier:
How are we doing with this?

baumeier:
Something does not seem right with this yet. I get 

```
Initializing calculator
... neighborlist
1 frames in statefile, Ids are: 0
Starting at frame 0
Evaluating frame 0
Import MD Topology (i.e. frame 0) from state.hdf5
....
... neighborlist  Using 1 threads
Evaluating 12 segments for neighborlist. 5038 segments are not taken into account as specified
The following segments are used in the neigborlist creation
	 ss
 ... ... Evaluating
0%   10   20   30   40   50   60   70   80   90   100%
|----|----|----|----|----|----|----|----|----|----|
terminate called after throwing an instance of 'std::runtime_error'
  what():  Cutoff is larger than half the box size. Maximum allowed cutoff is 679.7
```

with something like this

```
<?xml version="1.0"?>
<options>
  <neighborlist help="Determines neighbouring pairs for exciton transport" section="sec:neighborlist">
    <segments>
      <type bla="segment name1 segment_name2">ss ss</type>
      <cutoff unit="nm">0.3</cutoff>
    </segments>
    <cutoff_type help="method to define the cutoff" default="constant" choices="constant,segmentname">segmentname</cutoff_type>
    <constant help="cutoff for all other types of pairs if the constant method is used" unit="nm" default="1.5" choices\
="float+"/>
    <use_exciton_cutoff help="Use exciton cutoff" default="false" choices="bool"/>
    <exciton_cutoff help="value" units="nm"/>
  </neighborlist>
</options>
```

felipeZ:
@JensWehner  I have removed the exists that you told me about :)

share/xtp/xml/molpol.xml
```diff
@@ -7,8 +7,10 @@
       <max_iter help="Maximum number of iterations" default="500"/>
       <exp_damp help="Thole sharpness parameter" default="0.39"/>
     </options_polar>
-    <qmpackage help="DFT package polarization calculation was run with" default="orca" choices="orca"/>
-    <logfile help="logfile of DFTpackage to read Target polarisability from" default="votca.orb"/>
+    <mode help="Read the polar target from a file or a qmpackage logfile" choices="target,qmpackage"/>
+    <target help="should have this format: pxx pxy pxz pyy pyz pzz"/>
```
JensWehner, line 11:
```suggestion
    <target_polarisability help="should have this format: pxx pxy pxz pyy pyz pzz"/>
```

share/xtp/xml/neighborlist.xml
```diff
@@ -1,6 +1,11 @@
 <?xml version="1.0"?>
 <options>
   <neighborlist help="Determines neighbouring pairs for exciton transport" section="sec:neighborlist">
-    <constant help="cutoff for all other types of pairs" unit="nm" default="1.5" choices="float+"/>
+    <type help="segments names if the segmentname method is used"/>
```
JensWehner, line 4:
```suggestion
    <segmentname help="segments names if the segmentname method is used"/>
```

share/xtp/xml/neighborlist.xml
```diff
@@ -1,6 +1,11 @@
...
+    <cutoff help="cutoff value if the segmentname method is used" unit="nm"/>
+    <cutoff_type help="method to define the cutoff" default="constant" choices="constant,segmentname"/>
+    <constant help="cutoff for all other types of pairs if the constant method is used" unit="nm" default="1.5" choices="float+"/>
+    <use_exciton_cutoff help="Use exciton cutoff" default="false" choices="bool"/>
+    <exciton_cutoff help="value" units="nm"/>
```
JensWehner, line 9:
```suggestion
    <exciton_cutoff help="cutoff for classical exciton transition charge treatment" units="nm"/>
```

src/libxtp/tools/molpol.cc
```diff
@@ -44,20 +44,10 @@ void MolPol::Initialize(const tools::Property& user_options) {
...
-  if (!target_exists && !qmpackage_exists) {
-    throw std::runtime_error(
-        "You have to define a polar targer <target> or a or qmpackage logfile");
-  }
-
-  if (target_exists) {
+  // polar targer or qmpackage logfile
+  const std::string& mode = options.get("mode").as<std::string>();
 
+  if (mode == "target") {
```
JensWehner, line 50: "target" is not a good name for the mode. Maybe "explicit polarisation"

felipeZ: I think that `file` is more meaningful because the help message says *Read the polar target from a file or a qmpackage logfile*

JensWehner: yeah the help message is wrong, "Read the target from a qmpackage logfile or enter it exlicitly"

JensWehner: Which was most likely my doing

share/xtp/xml/neighborlist.xml
felipeZ: @baumeier I am introducing this change, just to let you know because you are using this feature

PR #521, opened 2020-07-29
added more checks to the dft_parse_part

junghans:


junghans:
@votca-bot changelog: add more checks to the dft_parse_part

junghans:
@baumeier this is for stable.

Issue #67, opened 2018-01-04
Add support for molden as an output format


JensWehner:
maybe a new calculator

http://www.cmbi.ru.nl/molden/molden_format.html

PR #524, opened 2020-07-31
Molden parser

rubengerritsen:
A new tool to convert molden files to .orb files.

Since molden files do not store any information about exact exchange, we need to use the exact exchange based on the functional name. That is why the gwbse object had to change as well. Now it checks at the start if the exact exchange of the functional matches that in the .orb file. If not, it will no longer abort the run, but display a warning and continue with the exact exchange of the functional provided to gwbse via the option file.

I was also wondering if I should include some proof of the fact that the tool is correct and behaves as expected. If so let me know how.


src/libxtp/tools/mol2orb.cc
```diff
@@ -0,0 +1,272 @@
...
+    double x, y, z;
+    int atom_id;
+    std::string junk;
+    std::string atom_type;
+    iss >> atom_type >> atom_id >> junk >> x >> y >> z;
+    atom_id =
+        atom_id - 1;  // molden uses indexing from 1, we use indexing from 0
+
+    // Add data to orbitals object
+    using Atom =
```
JensWehner, line 68: yeah in this case it is simply a QMAtom

src/libxtp/tools/mol2orb.cc
```diff
@@ -0,0 +1,272 @@
...
+    Eigen::Vector3d pos(x, y, z);
+    if (units == "Angs") {
+      pos = tools::conv::ang2bohr * pos;
+    }
+    mol.push_back(Atom(atom_id, atom_type, pos));
+  }
+  return "";
+}
+
+inline std::vector<std::array<int, 2>> Mol2Orb::getTwoCycles(Index numFunc) {
```
JensWehner, line 79: why inline and it needs a better name

src/libxtp/tools/mol2orb.cc
```diff
@@ -0,0 +1,272 @@
...
+      v.row(currentFunction + cycle[0]).swap(v.row(currentFunction + cycle[1]));
+    }
+    currentFunction += nrOfFunctions;
+  }
+
+  // Multiply by multiplier
+  for (Index i = 0; i < v.cols(); i++) {
+    v.row(i) = multiplier[i] * v.row(i);
+  }
+}
```
JensWehner, line 124: okay this is basically the same stuff that I merged into qmpackage. Maybe we should extract it into a separate reorder class?

src/libxtp/tools/mol2orb.cc
```diff
@@ -0,0 +1,272 @@
...
+    iss >> tempStr >> tempDouble;
+    orbitals.MOs().eigenvalues()[i] = tempDouble;
+    // spin channel line
+    std::getline(input_file, line);
+    iss.str(line);
+    iss.clear();
+    iss >> tempStr >> tempStr;
+    if (tempStr == "Beta") {
+      XTP_LOG(Log::error, _log)
+          << "Open shell systems are currently not supported" << std::flush;
```
JensWehner, line 160: why not throw an exception, because basically this finishes the program

src/libxtp/tools/mol2orb.cc
```diff
@@ -0,0 +1,272 @@
...
+        "Basis size not set, atoms were not parsed first.");
+  }
+  orbitals.MOs().eigenvalues().resize(basis_size);
+  orbitals.MOs().eigenvectors().resize(basis_size, basis_size);
+
+  // actual parsing
+  std::string line;
+  std::string tempStr;
+  double tempDouble;
+  int tempInt;
```
JensWehner, line 143: if possible do not use `int` but `Index`

src/libxtp/tools/mol2orb.cc
```diff
@@ -0,0 +1,272 @@
...
+
+  Orbitals orbitals;
+
+  // Open the file
+  std::ifstream input_file(_moldenfile);
+  // Check if succesfull
+  if (input_file.fail()) {
+    XTP_LOG(Log::error, _log) << "Could not open " << _moldenfile << std::flush;
+    return false;
+  }
```
JensWehner, line 204: maybe exceptions?

src/libxtp/tools/mol2orb.cc
```diff
@@ -0,0 +1,272 @@
...
+      boost::trim(units);
+      XTP_LOG(Log::error, _log)
+          << "Reading atoms using " << units << " units." << std::flush;
+      line = readAtoms(mol, units, input_file);
+      setupBasis(orbitals);
+    } else if (sectionType == "GTO") {
+      XTP_LOG(Log::error, _log)
+          << "Basisset specification is ignored." << std::flush;
+      XTP_LOG(Log::error, _log)
+          << "Basissets are specified via the mol2orb.xml options file."
```
JensWehner, line 238: why not read the basisset from molden as well? What are the dangers of doing it like this?

src/libxtp/tools/mol2orb.h
```diff
@@ -0,0 +1,79 @@
...
+
+namespace votca {
+namespace xtp {
+class Orbitals;
+
+class Mol2Orb : public QMTool {
+ public:
+  Mol2Orb() = default;
+
+  ~Mol2Orb() override = default;
```
JensWehner, line 23: you can turn override into final

src/libxtp/tools/mol2orb.h
```diff
@@ -0,0 +1,79 @@
...
+#include <votca/tools/constants.h>
+
+// Local VOTCA includes
+#include "votca/xtp/logger.h"
+#include "votca/xtp/qmtool.h"
+#include <votca/xtp/orbitals.h>
+
+namespace votca {
+namespace xtp {
+class Orbitals;
```
JensWehner, line 17: you do not need to define Orbitals here, because you included it above anyway.

src/libxtp/tools/mol2orb.h
```diff
@@ -0,0 +1,79 @@
...
+    std::array<int, 2>{1, 2},
+    std::array<int, 2>{3, 4},
+    std::array<int, 2>{5, 6}
+  };
+  std::vector<std::array<int, 2>> _twoCyclesG{
+    std::array<int, 2>{1, 2},
+    std::array<int, 2>{3, 4},
+    std::array<int, 2>{5, 6},
+    std::array<int, 2>{7, 8}
+  };
```
JensWehner, line 57: I do not understand this

src/libxtp/tools/mol2orb.h
```diff
@@ -0,0 +1,79 @@
...
+  std::string _basisset_name;
+  std::string _aux_basisset_name;
+  AOBasis _basis;
+  AOBasis _auxbasis;
+  Logger _log;
+
+  inline std::string readAtoms(QMMolecule& mol, std::string units,
+                               std::ifstream& input_file) const;
+  inline std::string readMOs(Orbitals& orbitals, std::ifstream& input_file);
+  inline std::vector<std::array<int, 2>> getTwoCycles(Index numFunc);
```
JensWehner, line 71: why inline?

JensWehner: why do these functions return a std::string?

src/libxtp/tools/mol2orb.cc
```diff
@@ -0,0 +1,272 @@
...
+
+  // fill DFT AO basis by going through all atoms
+  _basis.Fill(bs, orbitals.QMAtoms());
+  orbitals.setDFTbasisName(_basisset_name);
+  XTP_LOG(Log::error, _log)
+      << TimeStamp() << " Basisset size: " << _basis.AOBasisSize()
+      << std::flush;
+  orbitals.setBasisSetSize(_basis.AOBasisSize());
+  orbitals.setAuxbasisName(_aux_basisset_name);
+}
```
JensWehner, line 44: The name is not Good, because what you are actually doing is adding the basisset information to orbitals.

src/libxtp/tools/mol2orb.cc
```diff
@@ -0,0 +1,272 @@
...
+    // Add data to orbitals object
+    using Atom =
+        typename std::iterator_traits<QMMolecule::iterator>::value_type;
+    Eigen::Vector3d pos(x, y, z);
+    if (units == "Angs") {
+      pos = tools::conv::ang2bohr * pos;
+    }
+    mol.push_back(Atom(atom_id, atom_type, pos));
+  }
+  return "";
```
JensWehner, line 76: why not `void` then

src/libxtp/tools/mol2orb.cc
JensWehner: I do not understand what it does.

JensWehner:
Concerning proof, we should parse a molden file and see if it gets the correct data from it. the orca test has stuff like that

src/libxtp/tools/mol2orb.h
rubengerritsen: Inline, it can do without, I will remove it.

The reason they return a string is that in a molden file there is no number of atoms provided and there aren't any empty lines between sections. So you know that you've read all the atoms as soon as you read the line with the specification of the next section (same for the MOs). So what it returns is the next section type to read.

JensWehner: Okay I think that warrants a comment. I did not know that molden is so unstructured. Then it makes sense.

rubengerritsen:
> Concerning proof, we should parse a molden file and see if it gets the correct data from it. the orca test has stuff like that

I have tested this and it does, the orb file it generates can then be used for a gwbse calculation and it matches the input>orca>parse>gwbse outcome. Also the orb files generated by "parse" in dftgwbse and the molden parser are the same up to the digit for 10^(-6) and almost always up to the 10^(-8) digit. So there isn't much (maybe even none) accuracy loss by using molden files instead of the orca binaries.

I was thinking more of a unit test or something like that, but maybe that is overkill for a parser.

JensWehner:
I am working on something to make parsing files for tests easier. So I would like to have a unit test for the parsers, because if they break it is really hard to debug it.

JensWehner:
@votca-bot make format

JensWehner:
@votca-bot format

JensWehner:
@votca-bot changelog: Add a molden parser to xtp

src/libxtp/tools/mol2orb.cc
rubengerritsen: This is just in case, in principle it will never get here but the parser will find a new section to read with this piece of code 

``` 
if (line == "" || line[0] == '[') {
      return line;
    }
```

and return that section, in case that it is the last section in the molden file it returns an empty string.

JensWehner:
Do you also want to make a molden writer?

rubengerritsen:
At the moment I see no use for a molden writer, maybe in the future, but for now I just want to use the parser to start experimenting with the spin in the orbitals object.

Some info on these two-cycles will follow shortly, I wikipediaed it and normal people call them transpositions, so maybe that helps a bit already.

rubengerritsen:
The promised info on the 2-cycles (transpositions) and also why the reordering in this PR is not a carbon copy of the qmpackage.

In the qmpackage reordering happens as follows:

1. in the specific package (e.g. orca.h) relative shifts of orbitals are provided in the _reorder variable
2. from these a reordering vector is created  for one shell in `QMPackage::getReorderShell`
3. Then for all shells these vectors are combined in one long reordering vector `QMPackage::getReorderVector`
4. Then in `QMPackage::ReorderMOs` this long reorder vector is converted back to the relative shifts or actually transpositions. with this bit of code

```
// actual swapping of coefficients
  for (Index s = 1, d; s < (Index)order.size(); ++s) {
    for (d = order[s]; d < s; d = order[d]) {
      ;
    }
    if (d == s) {
      while (d = order[d], d != s) {
        v.row(s).swap(v.row(d));
      }
    }
  }
```

Since I didn't want to make a seperate qmpackage for molden files just to be able to reorder, I propose a different way of reordering in this pull request.

My plan:

Provide the transpositions directly and apply them immediately without building a reordering vector in between

```
  for (const AOShell& shell : _basis) {
    // Make multiplier vector
    .
    .
    .

    // reorder
    nrOfFunctions = shell.getNumFunc();
    for (auto& transposition : getTranspositions(nrOfFunctions)) {
      v.row(currentFunction + transposition[0]).swap(v.row(currentFunction + transposition[1]));
    }
    currentFunction += nrOfFunctions;
  }
```

rubengerritsen:
I find this 

```
std::vector<std::array<int, 2>> _TranspositionsS{};
  std::vector<std::array<int, 2>> _TranspositionsP{
    std::array<int, 2>{0, 2}
  };
  std::vector<std::array<int, 2>> _TranspositionsD{
    std::array<int, 2>{1, 2},
    std::array<int, 2>{3, 4}
    };
  std::vector<std::array<int, 2>> _TranspositionsF{
    std::array<int, 2>{1, 2},
    std::array<int, 2>{3, 4},
    std::array<int, 2>{5, 6}
  };
  std::vector<std::array<int, 2>> _TranspositionsG{
    std::array<int, 2>{1, 2},
    std::array<int, 2>{3, 4},
    std::array<int, 2>{5, 6},
    std::array<int, 2>{7, 8}
  };
```

particularly ugly by the way, but I couldn't think of a way to store the pairs differently, so suggestions are welcome

rubengerritsen:
When it comes to also parsing the basisset, I am wondering how that should be done, since votca (for as far as I know) doesn't support external basissets. No basisset is stored, it is just the name in the .orb files and then the basisset is constructed as it is needed.

JensWehner:
Good points you are making here. 

So a) reordering the individual shells would be a good idea in general. So basically you disassembled the reorder vector into the individual exchanges. 

b) I would like to keep all the reordering in one place, because otherwise it will be a pain to ever change it/maintain it. 

c) A molden writer, would be nice for output visualisation in general and for interfacing with other programs. See https://github.com/votca/xtp/issues/67

d) Good question, I actually would like to store the `aobasis` inside the HDF file as well instead of just the string. I never came around to doing it though.

rubengerritsen:
I have changed this into a draft for now because of a new class I have written, namely "OrbReorder" that deals with points (a) and (b). Let me know what you think. 

I will also write a molden writer, can't be to hard and I have made an issue for (d).

junghans:
@votca-bot changelog: Add a molden parser to xtp

junghans:
@votca-bot changelog: Add a molden parser to xtp

junghans:
@votca-bot format

rubengerritsen:
@votca-bot format

JensWehner:
@junghans we had a conversation about that and tests are coming. Unfortunately some of our older code in votca is not a very good example of how to write testable code.

junghans:
> @junghans we had a conversation about that and tests are coming. Unfortunately some of our older code in votca is not a very good example of how to write testable code.

Thanks, just saw the low coverage on the new code.

JensWehner:
@rubengerritsen as a small sidenote, [libint2](https://github.com/evaleev/libint/blob/7a9ef4b51ff773f30bd7491aea333c1ab6dac304/include/libint2/lcao/molden.h), a library which we are planning to adapt has a molden writer I thing.

PR #526, opened 2020-08-03
fix case no statefile exists yet and improved warnings

JensWehner:
I changed the locking behaviour earlier to make them more robust, but I forgot the case if no state file exists yet.


JensWehner:
@votca-bot changelog: write an empty state file if it does not exist yet

PR #528, opened 2020-08-03
Fix changelog

junghans:
Fixup for #526 , still need to investigate why it actually happened.


junghans:
The problem was there were spaces in the empty line between the version sections, so the last line was detected incorrectly, fixed here: https://github.com/votca/actions/commit/fe762fe7e0fe645041896f92e8aa87cc607ffd6e

PR #529, opened 2020-08-03
Unit test rework

JensWehner:
Like in tools and csg large matrices and other larger files are not stored in the source code anymore but instead are read in from files to reduce compile times.


JensWehner:
@junghans please review! I deleted the large memory test thing, because the compilation of the test should not require large memory anymore, because everything is done at runtime.

JensWehner:
please do not delete branch, after merging because there are more tests to refactor but I think this already warrants a merge.

JensWehner:
@votca-bot changelog: Unittests can read input and reference values from files

JensWehner:
@votca-bot format

rubengerritsen:
@votca-bot format

rubengerritsen:
@votca-bot format

PR #530, opened 2020-08-04
cmake: fix parallel units tests

junghans:
All the unit tests write the same files, hence running unit tests in parallel fails, so let's put each of the tests in it own dir.


JensWehner:
I sort of wanted to fix this issue by reading in most of the files. But this would of course also work.

junghans:
> I sort of wanted to fix this issue by reading in most of the files. But this would of course also work.

Yeah, this is for stable, you can undo this change in master if you like.

PR #536, opened 2020-08-06
reimplemented the pair parsing

JensWehner:
This is not a fix yet, we have a problem with defaults and updating with user options, because sometimes multiple options with the same tag are required. So this bug is actually in tools

@felipeZ


JensWehner:
not a useful fix

Issue #538, opened 2020-08-19
Getting Rid of combined shells


JensWehner:
At the Moment VOTCA supports shells of the type "SPD" or "PDFG", 

I want to remove that feature and only allow shells of the type "S" "P" "D" etc... .

Reasons:

a) Most basissets we use do not have that kind of shells
b) Most basissets are contracted and it allows us to optimize that path a lot more. 
c) It makes implementing new features a lot easier because this special case does not have to be treated anymore. Easier for new comers
d) Libint and Orca do no have it and even gaussian only supports SP shells, for which the performance gain is small.
e) We could indicate shells not by a `char` or a `string` but simply by `L` which is a lot nicer. 

Cons:

It would hurt the most for aux-basissets that were constructed with our `make_auxbasis` scripts, but even there I think the performance gain from maybe switching to libint at some point might be bigger. 

In general as we store integrals in memory at the time, this whole thing will not be critical. 

@baumeier any thoughts?
@felipeZ thoughts?
@rubengerritsen might be helpful for your stuff as well

Possible implementation workflow:
Basisset splits up combined shells when it reads them in, into single shells. This is fully backward compatible.
From there we can bit by bit remove all other places where combined shells are used.

felipeZ:
I think that if we plan to move to `libint` in the future it is a good think to do

baumeier:
No objections.

JensWehner:
Actually our whole .xml basisset thing is really dangerous, because we assume that the shells are ordered in the xml,, while the xml format does not guarantee it. So basically every shell should have an index.

i.e. 
<element>
<shell "number"=0>
...
</shell>
<shell "number"=1>
...
</shell>
</element>

PR #539, opened 2020-08-19
Single shells

JensWehner:
implements https://github.com/votca/xtp/issues/538

Shells do now have a single enum which tells their L value.


JensWehner:
@votca-bot changelog: xtp internally uses only single shells.

src/libxtp/basisset.cc
```diff
@@ -26,169 +26,101 @@
...
+      return "F";
+    case L::G:
+      return "G";
+    case L::H:
+      return "H";
+    case L::I:
+      return "I";
   }
-  return nbf;
+  return "";
```
felipeZ, line 76: Why not raising an error here?

src/libxtp/basisset.cc
```diff
@@ -26,169 +26,101 @@
...
+      return 9;
+    case L::G:
+      return 16;
+    case L::H:
+      return 25;
+    case L::I:
+      return 36;
   }
-  return nbf;
+  return -1;
```
felipeZ, line 96: Is there a case where this `-1` could be a problem?

felipeZ (approved):
Great work:
1. It is more clear
2. some duplicated functionality has been removed
3. This changes pave the way to introduced LibInt in the future

src/libxtp/basisset.cc
```diff
@@ -26,169 +26,101 @@
...
+    case L::D:
+      return "D";
+    case L::F:
+      return "F";
+    case L::G:
+      return "G";
+    case L::H:
+      return "H";
+    case L::I:
+      return "I";
```
rubengerritsen, line 74: maybe add a default that throws an error, same for the switches that follow this one

rubengerritsen (commented):
Very nice, it makes the code way easier to understand

src/libxtp/basisset.cc
JensWehner: If I am not mistaken you will get a compiler warning if you do not treat the whole enum so this cannot happen anymore.

src/libxtp/basisset.cc
JensWehner: I think it should never happen.

src/libxtp/basisset.cc
JensWehner: yes I agree. I hope that it actually produces a compiler warning if you do not treat all the enum cases, so that you do not need a default.

JensWehner: https://stackoverflow.com/questions/37254496/c-warning-enumeration-value-not-handled-in-switch-wswitch actually without the default you get a compiler warning, which I prefer.

src/libxtp/basisset.cc
JensWehner: https://stackoverflow.com/questions/37254496/c-warning-enumeration-value-not-handled-in-switch-wswitch

src/libxtp/basisset.cc
rubengerritsen: Me to, didn't know about that

src/libxtp/basisset.cc
JensWehner: I am not sure I did it to get rid of the compiler warning

src/libxtp/basisset.cc
felipeZ: thanks, I didn't know!

PR #548, opened 2020-09-01
CDA implementation for the self energy

gtirimbo:
This is a working implementation of the Contour Deformation Approach (CDA) for the self-energy correlation part. 
A unit test is provided.


include/votca/xtp/gw.h
```diff
@@ -63,6 +63,9 @@ class GW {
     double qp_grid_spacing;  // Spacing of grid points in Ha
     Index gw_mixing_order;   // mixing order
     double gw_mixing_alpha;  //  mixing alpha, also linear mixing
+    std::string quadrature_scheme;  // Kind of Gaussian-quadrature scheme to use
+    Index order = 12;  // only needed for complex integration sigma default:12
+    double alpha = 0.001;
```
JensWehner, line 68: what is alpha?

src/libxtp/gaussian_quadrature.cc
```diff
@@ -0,0 +1,225 @@
...
+              << std::endl;
+  }
+}
+
+// This function calculates and stores inverses of the microscopic dielectric
+// matrix in a matrix vector
+void GaussianQuadrature::CalcDielInvVector(const RPA& rpa) {
+  _dielinv_matrices_r.resize(_opt.order);
+  double halfpi = std::acos(0.0);
+#pragma openmp parallel schedule(guided)
```
JensWehner, line 68: This only parallelizes up to the order which can be small. Is that the parallelisation scheme you want?

src/libxtp/gaussian_quadrature.cc
```diff
@@ -0,0 +1,225 @@
...
+  for (Index j = 0; j < _opt.order; j++) {
+    if (_opt.quadrature_scheme == "legendre") {
+      double newpoint = std::tan(halfpi * _quadpoints(j));
+      Eigen::MatrixXcd eps_inv_j =
+          rpa.calculate_epsilon_complex(0.0, newpoint).inverse();
```
JensWehner, line 73: You calculate it on the real axis, do you think we could use on of the other RPA methods here?

src/libxtp/gaussian_quadrature.cc
```diff
@@ -0,0 +1,225 @@
...
+      eps_inv_j.diagonal().array() -= 1.0;
+      _dielinv_matrices_r[j] = -eps_inv_j;
+    } else {
+      double newpoint = _quadpoints(j);
+      Eigen::MatrixXcd eps_inv_j =
+          rpa.calculate_epsilon_complex(0.0, newpoint).inverse();
+      eps_inv_j.diagonal().array() -= 1.0;
+      _dielinv_matrices_r[j] = -eps_inv_j;
```
JensWehner, line 88: these lines are the same for all methods so call them afterwards, or even better the different `quadrature` classes should have a function `Point()` or so which returns the correct `newpoint`

src/libxtp/gaussian_quadrature.cc
```diff
@@ -0,0 +1,225 @@
...
+double GaussianQuadrature::SigmaGQDiag(double frequency, Index gw_level,
+                                       double eta) const {
+  Index homo = _opt.homo - _opt.rpamin;
+  Index lumo = homo + 1;
+
+  const Eigen::MatrixXd& Imx = _Mmn[gw_level];
+  Eigen::ArrayXcd DeltaE = frequency - _energies.array();
+  std::complex<double> result = std::complex<double>(0.0, 0.0);
+  if (_opt.quadrature_scheme == "laguerre") {
+    // This work for integration limits a = 0 b = +infty
```
JensWehner, line 102: What is `this`?

src/libxtp/gaussian_quadrature.cc
```diff
@@ -0,0 +1,225 @@
...
+
+    for (Index k = 0; k < lumo; ++k) {
+      DeltaE(k) += std::complex<double>(0.0, 1.0 * eta);
+    }
+    for (Index k = lumo; k < _energies.size(); ++k) {
+      DeltaE(k) += std::complex<double>(0.0, -1.0 * eta);
+    }
```
JensWehner, line 109: Delta
```suggestion
  DeltaE.head(occ).imag()=eta
  DeltaE.tail(unocc).imag()=-eta
```

include/votca/xtp/gw.h
gtirimbo: https://aip.scitation.org/doi/10.1063/1.5144277
Equation 28 and 29

src/libxtp/gaussian_quadrature.cc
```diff
@@ -0,0 +1,225 @@
...
+
+    for (Index j = 0; j < _opt.order; ++j) {
+      double newpoint = _quadpoints(j);
+      double newweight = _quadadaptedweights(j);
+      Eigen::VectorXcd denominator1 =
+          (1.0) / (DeltaE + std::complex<double>(0.0, newpoint));
```
JensWehner, line 115:
```suggestion
         (DeltaE + std::complex<double>(0.0, newpoint)).cwiseinverse()
```

Or something similar

src/libxtp/gaussian_quadrature.cc
```diff
@@ -0,0 +1,225 @@
...
+      Eigen::MatrixXcd Amx = denominator1.asDiagonal() * Imx;
+      Eigen::MatrixXcd Cmx = Imx * (_dielinv_matrices_r[j]);
```
JensWehner, line 117: do you really need this intermediate values. I would just put the whole equation in the next line. Eigen will optimise the unnecessary temp objects away or declare them `auto` instead of Eigen types

src/libxtp/gaussian_quadrature.cc
```diff
@@ -0,0 +1,225 @@
...
+      std::complex<double> value1 = (Cmx.cwiseProduct(Amx)).sum();
+      Eigen::VectorXcd denominator2 =
+          (1.0) / (DeltaE + std::complex<double>(0.0, -newpoint));
+      Eigen::MatrixXcd Dmx = denominator2.asDiagonal() * Imx;
+      Eigen::MatrixXcd Emx = Imx * (_dielinv_matrices_r[j].conjugate());
```
JensWehner, line 122: same here.

src/libxtp/gaussian_quadrature.cc
```diff
@@ -0,0 +1,225 @@
...
+
+    for (Index j = 0; j < _opt.order; ++j) {
+
+      Eigen::VectorXcd denominator =
+          (1.0) / (DeltaE + std::complex<double>(0.0, _quadpoints(j)));
+      Eigen::MatrixXcd Amx = denominator.asDiagonal() * Imx;
+      Eigen::MatrixXcd Cmx = Imx * (_dielinv_matrices_r[j]);
+      std::complex<double> value = (Cmx.cwiseProduct(Amx)).sum();
+      result += _quadadaptedweights(j) * value;
+    }
```
JensWehner, line 179: this is all basically the same as before. Please refactor but either making the individual quadrature schemes, derived classes of some `quadrature_scheme_base` or so. If you need help let me know.

src/libxtp/gaussian_quadrature.cc
```diff
@@ -0,0 +1,225 @@
...
+      double newweight = _quadadaptedweights(j);
+      Eigen::VectorXcd denominator1 =
+          (num) / (DeltaE + std::complex<double>(0.0, newpoint));
+      Eigen::MatrixXcd Amx = denominator1.asDiagonal() * Imx;
+      Eigen::MatrixXcd Cmx = Imx * (_dielinv_matrices_r[j]);
+      std::complex<double> value1 = (Cmx.cwiseProduct(Amx)).sum();
+      result += newweight * (value1);
+    }
+
+    result *= 0.5;
```
JensWehner, line 217: see above

src/libxtp/gaussian_quadrature.cc
```diff
@@ -0,0 +1,225 @@
...
+  }
+
+  return result.real() / (tools::conv::Pi);
```
JensWehner, line 220: here you used Pi, nice

src/libxtp/gwbse/gwbse.cc
```diff
@@ -358,6 +358,23 @@ void GWBSE::Initialize(tools::Property& options) {
...
+          key + ".quadrature_scheme", _gwopt.quadrature_scheme);
+  XTP_LOG(Log::error, *_pLog)
+      << " Integration scheme Quadrature: " << _gwopt.quadrature_scheme
+      << flush;
+  _gwopt.alpha = options.ifExistsReturnElseReturnDefault<double>(key + ".alpha",
+                                                                 _gwopt.alpha);
+  XTP_LOG(Log::error, *_pLog) << " Alpha value: " << _gwopt.alpha << flush;
+
+  _gwopt.eta =
+      options.ifExistsReturnElseReturnDefault<double>(key + ".eta", _gwopt.eta);
```
JensWehner, line 376: defaults are now given by xml so please remove the `Return default` cases

src/libxtp/gwbse/rpa.cc
```diff
@@ -91,6 +91,46 @@ Eigen::MatrixXd RPA::calculate_epsilon(double frequency) const {
 template Eigen::MatrixXd RPA::calculate_epsilon<true>(double frequency) const;
 template Eigen::MatrixXd RPA::calculate_epsilon<false>(double frequency) const;
 
+Eigen::MatrixXcd RPA::calculate_epsilon_complex(double frequency_r,double frequency_i) const {
+  const Index size = _Mmn.auxsize();
+  std::vector<Eigen::MatrixXcd> thread_result = std::vector<Eigen::MatrixXcd>(
+      OPENMP::getMaxThreads(), Eigen::MatrixXcd::Zero(size, size));
```
JensWehner, line 97: we use custom reductors now, please add one to `eigen.h`

src/libxtp/gwbse/rpa.cc
```diff
@@ -91,6 +91,46 @@ Eigen::MatrixXd RPA::calculate_epsilon(double frequency) const {
 template Eigen::MatrixXd RPA::calculate_epsilon<true>(double frequency) const;
 template Eigen::MatrixXd RPA::calculate_epsilon<false>(double frequency) const;
 
+Eigen::MatrixXcd RPA::calculate_epsilon_complex(double frequency_r,double frequency_i) const {
```
JensWehner, line 94: Use a `std::complex<double>` instead of two `double`

src/libxtp/gwbse/rpa.cc
```diff
@@ -91,6 +91,46 @@ Eigen::MatrixXd RPA::calculate_epsilon(double frequency) const {
...
+  std::vector<Eigen::MatrixXcd> thread_result = std::vector<Eigen::MatrixXcd>(
+      OPENMP::getMaxThreads(), Eigen::MatrixXcd::Zero(size, size));
+  const Index lumo = _homo + 1;
+  const Index n_occ = lumo - _rpamin;
+  const Index n_unocc = _rpamax - lumo + 1;
+
+  std::complex<double> frequency(frequency_r, frequency_i);
+  std::complex<double> eta(0.0, _eta);  
+
+#pragma omp parallel for
```
JensWehner, line 105:
```suggestion
#pragma omp parallel for schedule(guided)
```

src/libxtp/gwbse/sigma_cda.cc
```diff
@@ -0,0 +1,120 @@
...
+// This calculates eps^-1 (inverse of the dielectric function) for complex
+// frequencies of the kind omega = delta + i*eta
+double Sigma_CDA::CalcDiagContribution(Eigen::RowVectorXd Imx_row, double delta,
+                                      double eta) const {
+  Eigen::MatrixXcd DielMxInv =
+      _rpa.calculate_epsilon_complex(delta, eta).inverse();
+  DielMxInv.diagonal().array() -= 1.0;
+  std::complex<double> lambda =
+      ((Imx_row * DielMxInv).cwiseProduct(Imx_row)).sum();
+  return lambda.real();
```
JensWehner, line 49:
```suggestion
 return  ((Imx_row * DielMxInv).cwiseProduct(Imx_row)).sum().real();
```

src/libxtp/gwbse/sigma_cda.cc
```diff
@@ -0,0 +1,120 @@
...
+  double factor = 0.0;
+
+  if (e_f < e_m && e_m < frequency) {
+    factor = 1.0;
+  } else if (e_f > e_m && e_m > frequency) {
+    factor = -1.0;
+  } else if (e_m - frequency == 0.0 && e_f > e_m) {
+    factor = -0.5;
+  } else if (e_m - frequency == 0.0 && e_f < e_m) {
+    factor = 0.5;
```
JensWehner, line 63: comparing a double to an exact value is dangerous, are you sure this is you want and not a `std::abs(e_m-frequency)<tolerance` ? This will give different results depending on the architecture

src/libxtp/gwbse/sigma_cda.cc
```diff
@@ -0,0 +1,120 @@
...
+  Index lumo = homo + 1;
+  double fermi_rpa = (rpa_energies(lumo) + rpa_energies(homo)) / 2.0;
+
+  const Eigen::MatrixXd& Imx = _Mmn[gw_level];
+  for (Index i = 0; i < rpatotal; ++i) {
+    double delta = std::abs(rpa_energies(i) - frequency);
+    double factor = CalcResiduePrefactor(fermi_rpa, rpa_energies(i), frequency);
+
+    // just sum contributing terms
+    if (std::abs(factor) > 0.1) {
```
JensWehner, line 84: what is 0.1, where does it come from?

src/libxtp/gwbse/sigma_cda.cc
```diff
@@ -0,0 +1,120 @@
...
+
+  const Eigen::VectorXd& RPAenergies = _rpa.getRPAInputEnergies();
+
+  double sigma_c_residue =
+      CalcResidueContribution(RPAenergies, frequency, gw_level);
+
+  double sigma_c_integral = _gq.SigmaGQDiag(frequency, gw_level, _opt.alpha);
+
+  return sigma_c_residue + sigma_c_integral;
+}
```
JensWehner, line 102: remove all the white lines

src/libxtp/gwbse/sigma_cda.cc
```diff
@@ -0,0 +1,120 @@
...
+
+double Sigma_CDA::CalcDiagContributionValue_alpha(Eigen::RowVectorXd Imx_row,
+                                                 double delta,
+                                                 double alpha) const {
+  Eigen::MatrixXcd R = _rpa.calculate_epsilon_complex(0.0, 0.0).inverse();
+
+  R.diagonal().array() -= 1.0;
```
JensWehner, line 109: I do not know how often you have to calculate it, but this part is independent of the input arguments of this function. Do you maybe have to calculate this only once?

JensWehner (changes requested):
It would also be nice to have subtests for the classes inside the cda. Nice work overall. There are a couple of questions and a bit of code duplication.

include/votca/xtp/sigma_cda.h
```diff
@@ -0,0 +1,83 @@
...
+                                              double frequency) const {
+    double h = 1e-3;
+    double plus = CalcCorrelationDiagElement(gw_level, frequency + h);
+    double minus = CalcCorrelationDiagElement(gw_level, frequency - h);
+    return plus - minus / (2 * h);
+  }
+  // Calculates Sigma_c off-diagonal elements
+  double CalcCorrelationOffDiagElement(Index, Index, double, double) const {
+    return 0;
+  }
```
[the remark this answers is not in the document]
gtirimbo: Although It should be easy to include the off-diagonal part I would like to keep only the diagonal one for now if possible

src/libxtp/gaussian_quadrature.cc
gtirimbo: This is on the imaginary axis. 
I built this function because in the future we might need a more flexible function able to accept complex number (both real and imaginary part different from zero)

gtirimbo: Do you think in this way I slowed things down?

src/libxtp/gaussian_quadrature.cc
gtirimbo: for 'this' I meant the laguerre method. I will be more explicit

include/votca/xtp/sigma_cda.h
```diff
@@ -0,0 +1,83 @@
...
+ *
+ */
+
+#ifndef _VOTCA_XTP_SIGMA_CDA_H
+#define _VOTCA_XTP_SIGMA_CDA_H
+#include <complex>
+#include <votca/xtp/gaussian_quadrature.h>
+#include <votca/xtp/logger.h>
+#include <votca/xtp/rpa.h>
+#include <votca/xtp/sigma_base.h>
```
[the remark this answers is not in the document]
gtirimbo: For all of them?

include/votca/xtp/sigma_cda.h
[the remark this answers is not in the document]
JensWehner: Do you want to merge it into master yet?

src/libxtp/gaussian_quadrature.cc
JensWehner: On the imaginary axis the result is purely real/imagninary (I forgot which) so if you use a complex type there, it is 4x more expensive at least.

include/votca/xtp/sigma_cda.h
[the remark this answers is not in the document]
JensWehner: all the ones which come from xtp

src/libxtp/gwbse/sigma_cda.cc
gtirimbo: The prefactor can be 1,-1,0.5,-0.5 or 0. We avoid calculating the Diagonal contribution if the prefactor is 0. We want to calculate it for all the other cases.  
I picked 0.1 for this reason.

JensWehner: hmmm I see. a) your answer should be a comment in the code. b) You basically want to check if it is non-zero so, 1e-10 or something is better.

gtirimbo: Done both

include/votca/xtp/sigma_cda.h
```diff
@@ -0,0 +1,83 @@
...
+    double minus = CalcCorrelationDiagElement(gw_level, frequency - h);
+    return plus - minus / (2 * h);
+  }
+  // Calculates Sigma_c off-diagonal elements
+  double CalcCorrelationOffDiagElement(Index, Index, double, double) const {
+    return 0;
+  }
+  double CalcResiduePrefactor(double e_f, double e_m, double frequency) const;
+  double CalcResidueContribution(Eigen::VectorXd rpa_energies, double frequency,
+                                 Index gw_level) const;
```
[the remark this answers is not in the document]
gtirimbo: Ok I just put them private

src/libxtp/gaussian_quadrature.cc
gtirimbo: I inherited this from another implementation so I am not fond of it. What parallel scheme do you suggest to use?

include/votca/xtp/sigma_cda.h
```diff
@@ -0,0 +1,83 @@
...
+
+ private:
+  double CalcDiagContribution(Eigen::RowVectorXd Imx_row, double delta,
+                              double eta) const;
+
+  double CalcDiagContributionValue_alpha(Eigen::RowVectorXd Imx_row,
+                                         double delta, double alpha) const;
```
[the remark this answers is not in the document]
gtirimbo: what do you mean ?

src/libxtp/gauss_hermite_quadrature_constants.cc
```diff
@@ -0,0 +1,367 @@
...
+  if (!this->_filled_Points) {
+    this->FillPoints();
+    _filled_Points = true;
+  }
+  if (_map_points.count(order) == 0)
+    throw invalid_argument("Quadrature order " + std::to_string(order) +
+                           " not in {8,10,12,14,16,18,20 and 100}. Please "
+                           " select one of these numbers");
+  return _map_points.at(order);
+}
```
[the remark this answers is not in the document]
gtirimbo: Also here I think I am not following you :D

src/libxtp/gaussian_quadrature.cc
JensWehner: I have no idea at the moment. If you want a more fine-grained parallelism you probably have to go one level deeper in the code.

JensWehner: I am not sure how critical it is, how important this part is, but if `order=12` then you can only use upt to 12 threads to parallelize it.

include/votca/xtp/sigma_cda.h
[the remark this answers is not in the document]
JensWehner: This function takes `Imx_row` by value, i.e. makes a copy. Maybe just taking `Imx_row` by `const` Reference would be better 
```suggestion
  double CalcDiagContribution(const Eigen::RowVectorXd& Imx_row, double delta,
                              double eta) const;

  double CalcDiagContributionValue_alpha(const Eigen::RowVectorXd& Imx_row,
                                         double delta, double alpha) const;
```

include/votca/xtp/sigma_base.h
```diff
@@ -43,6 +43,9 @@ class Sigma_base {
     Index rpamin = 0;
     Index rpamax = 0;
     double eta = 1e-3;
+    std::string quadrature_scheme;  // Kind of Gaussian-quadrature scheme to use
+    Index order = 12;  // only needed for complex integration sigma default:12
+    double alpha = 0.1;  // only needed for complex integration of sigma: default 0.1
```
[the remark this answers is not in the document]
gtirimbo: I think I would like to keep alpha with a default value of 0.001 (I changed it wherever it was different). The reason is that very seldom we have the need to change it hence the default value.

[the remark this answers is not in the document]
JensWehner: yes, but you get the default value from the xml file and thus there should not be two default values.

src/libxtp/gwbse/rpa.cc
gtirimbo: Do you want to have just one complex number i.e
`Eigen::MatrixXcd RPA::calculate_epsilon_complex(std::complex<double> frequency) `
or do you want simply that double ---> std:complex \<double\>
` Eigen::MatrixXcd RPA::calculate_epsilon_complex(std::complex<double>double frequency_r,std::complex<double> frequency_i)  `

JensWehner: one complex number

src/libxtp/gwbse/sigma_cda.cc
gtirimbo: I will try this way

include/votca/xtp/gw.h
junghans: And unsued?

junghans:
@votca-bot copyright

junghans:
@votca-bot copyright

junghans:
@votca-bot copyright

junghans:
@votca-bot copyright

junghans:
@votca-bot copyright

junghans:
@votca-bot copyright

src/libxtp/gaussian_quadrature.cc
gtirimbo: Ok maybe we should go for a static schedule, right?

src/libxtp/gaussian_quadrature.cc
gtirimbo: I put them afterwards

include/votca/xtp/gw.h
gtirimbo: what?

gtirimbo: If you meant unused, you are right. It is something I inherited from a different implementation and doing some tests with the old implementation it seems not so important after all. I will remove it everywhere.

src/libxtp/gwbse/sigma_cda.cc
gtirimbo: Not needed because I removed the alpha dependencies

src/libxtp/gwbse/gwbse.cc
gtirimbo: Done

src/libxtp/gwbse/rpa.cc
gtirimbo: Tell me more

src/libxtp/gwbse/rpa.cc
gtirimbo: Done

src/libxtp/gauss_hermite_quadrature_constants.cc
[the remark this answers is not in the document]
JensWehner: So the `getPoints()` function cannot be called inside an `openmp` loop, because you could get race conditions there. So I think we should do the map stuff differently. I have to think about it. My suggestion was to put the `FillPoints()` into the constructor of this class and delete the `if(_filled_Points)` clause and the stuff inside.

src/libxtp/gaussian_quadrature.cc
JensWehner: It will not really help either. If you have more threads than `order` on your machine, the remaining threads will idle.

baumeier: There is also parallelization in `rpa.calculate_epsilon_complex()`, so what happens if we have a parallel region inside a parallel region? Also, is the `inverse()` call in eigen using a parallel MKL backend?

JensWehner: That is a good question. So I think in general nested parallelism is enabled. So we have a pool of threads, whose size is determined by `omp_set_num_thread`, if the first `omp_parallel_for` does not require all threads, it will use those threads further down the line.

JensWehner: Same is true for MKL, as long as it uses OPENMP for threading, which it does in votca.

baumeier: So it is pretty unlikely that having parallelization here is going to have much of an effect, because probably the parallel sum over occ states in RPA is more relevant anyways, I assume.

JensWehner: yes

src/libxtp/gaussian_quadrature.cc
gtirimbo: I changed both in gaussian_quadrature and sigma_cda so that the dielectric function is always real (hence inverting real matrices) even when complex argument are given

src/libxtp/gaussian_quadrature.cc
```diff
@@ -0,0 +1,180 @@
...
+      double num = halfpi / den;
+      Eigen::VectorXcd denominator = (DeltaE + newpoint).cwiseInverse();
+      std::complex<double> value =
+          ((Imx * (_dielinv_matrices_r[j]))
+               .cwiseProduct(denominator.asDiagonal() * Imx))
+              .sum();
+      result += num * _quadadaptedweights(j) * value;
+    }
+  }
+  return 0.5 * result.real() / (tools::conv::Pi);
```
JensWehner, line 175: I like the comments. In the code we have at least 3 times an `if` statement, which queries which quadrature scheme it is, this should be refactored into something like the sigma_base and its derived classes.

src/libxtp/gaussian_quadrature.cc
```diff
@@ -0,0 +1,180 @@
...
+              << std::endl;
+  }
+}
+
+// This function calculates and stores inverses of the microscopic dielectric
+// matrix in a matrix vector
+void GaussianQuadrature::CalcDielInvVector(const RPA& rpa) {
+  _dielinv_matrices_r.resize(_opt.order);
+  Eigen::MatrixXd eps_inv_j;  // Don't know if I have to specify dimension of
+                              // the matrix here
```
JensWehner, line 70:
```suggestion
  _dielinv_matrices_r.reserve(_opt.order);
```

src/libxtp/gwbse/rpa.cc
```diff
@@ -91,6 +91,74 @@ Eigen::MatrixXd RPA::calculate_epsilon(double frequency) const {
...
+  const Index n_occ = lumo - _rpamin;
+  const Index n_unocc = _rpamax - lumo + 1;
+
+#pragma omp parallel for schedule(dynamic) reduction(+ : result)
+  for (Index m_level = 0; m_level < n_occ; m_level++) {
+    const double qp_energy_m = _energies(m_level);
+    const Eigen::MatrixXd Mmn_RPA = _Mmn[m_level].bottomRows(n_unocc);
+    const Eigen::ArrayXd deltaE = _energies.tail(n_unocc).array() - qp_energy_m;
+
+    Eigen::VectorXd chi;
```
JensWehner, line 109: only define chi when you use it, i.e. further below

src/libxtp/gwbse/rpa.cc
JensWehner: If you look at the other RPA functions, they simply use a `reduce` keyword, for `MatrixXcd` this is not defined yet so you have to add that in the `eigen.h` header.

src/libxtp/gwbse/rpa.cc
```diff
@@ -91,6 +91,74 @@ Eigen::MatrixXd RPA::calculate_epsilon(double frequency) const {
...
+
+    Eigen::VectorXcd chi;
+    Eigen::ArrayXcd deltaEm = frequency - deltaE + eta;
+    Eigen::ArrayXcd deltaEp = frequency + deltaE - eta;
+
+    chi = deltaEm.cwiseInverse() - deltaEp.cwiseInverse();
+    Eigen::MatrixXcd tempresult =
+        Mmn_RPA.transpose() * chi.asDiagonal() * Mmn_RPA;
+
+    thread_result[OPENMP::getThreadId()] += tempresult;
```
JensWehner, line 151: see comment above

src/libxtp/gwbse/sigma_cda.cc
```diff
@@ -0,0 +1,102 @@
...
+  _gq.configure(opt, _rpa);
+}
+
+// This function is used in the calculation of the residues
+// This calculates eps^-1 (inverse of the dielectric function) for complex
+// frequencies of the kind omega = delta + i*eta
+double Sigma_CDA::CalcDiagContribution(const Eigen::RowVectorXd& Imx_row,
+                                       double delta, double eta) const {
+  std::complex<double> delta_eta(delta, eta);
+  Eigen::MatrixXd DielMxInv = _rpa.calculate_epsilon_r(delta_eta).inverse();
```
JensWehner, line 44: actually, you never have to explicitly calculate the inverse directly. All you are interested in is how the inverse acts on a vector. So solve the linear system instead,  so basically you want A-1*b, which you can get from solving Ax=b

src/libxtp/gwbse/sigma_cda.cc
```diff
@@ -0,0 +1,102 @@
...
+  opt.qpmin = _opt.qpmin;
+  opt.rpamin = _opt.rpamin;
+  opt.quadrature_scheme = _opt.quadrature_scheme;
+  _gq.configure(opt, _rpa);
+}
+
+// This function is used in the calculation of the residues
+// This calculates eps^-1 (inverse of the dielectric function) for complex
+// frequencies of the kind omega = delta + i*eta
+double Sigma_CDA::CalcDiagContribution(const Eigen::RowVectorXd& Imx_row,
```
JensWehner, line 41: I think this `.row()` is technically not a `RowVectorXd&`

src/libxtp/gwbse/sigma_cda.cc
```diff
@@ -0,0 +1,102 @@
...
+    factor = -1.0;
+  } else if (std::abs(e_m - frequency) < tolerance && e_f > e_m) {
+    factor = -0.5;
+  } else if (std::abs(e_m - frequency) < tolerance && e_f < e_m) {
+    factor = 0.5;
+  }
+  return factor;
+}
+
+double Sigma_CDA::CalcResidueContribution(Eigen::VectorXd rpa_energies,
```
JensWehner, line 65: no need for the copy of rpa_energies

src/libxtp/gwbse/rpa.cc
baumeier: Is this function needed at all any more?

src/libxtp/gwbse/sigma_cda.cc
```diff
@@ -0,0 +1,102 @@
...
+                                          double frequency,
+                                          Index gw_level) const {
+  Index rpatotal = rpa_energies.size();
+  double sigma_c = 0.0;
+  Index homo = _opt.homo - _opt.rpamin;
+  Index lumo = homo + 1;
+  double fermi_rpa = (rpa_energies(lumo) + rpa_energies(homo)) / 2.0;
+  const Eigen::MatrixXd& Imx = _Mmn[gw_level];
+  for (Index i = 0; i < rpatotal; ++i) {
+    double delta = std::abs(rpa_energies(i) - frequency);
```
JensWehner, line 75: move this into the if clause, otherwise it is not needed.

src/libxtp/gwbse/rpa.cc
baumeier: The whole `calculate_epsilon_complex` function is never used. Will remove.

src/libxtp/gwbse/rpa.cc
baumeier: see above

src/libxtp/gwbse/sigma_cda.cc
baumeier: What is it then? `MatrixXd&`?

src/libxtp/gwbse/sigma_cda.cc
baumeier: Not sure what you want to move here

src/libxtp/gwbse/sigma_cda.cc
baumeier: Any suggestion about the preferred eigen method for this?

src/libxtp/gwbse/sigma_cda.cc
JensWehner: `const Ref<const MatrixXd>&`

src/libxtp/gwbse/sigma_cda.cc
JensWehner: I would use the cheapest first so `llt()` 

https://eigen.tuxfamily.org/dox/group__TutorialLinearAlgebra.html

src/libxtp/gwbse/sigma_cda.cc
JensWehner: delta is only needed in the if clause further down

baumeier: No longer true ;) wait for it

src/libxtp/gwbse/sigma_cda.cc
baumeier: Is done in the residue contribution. Currently checking the performance of the different linear system solvers. According to https://eigen.tuxfamily.org/dox/group__DenseDecompositionBenchmark.html `PartialPivLU` should be the best choice in terms of speed.

src/libxtp/gwbse/sigma_cda.cc
baumeier: changed

src/libxtp/gwbse/sigma_cda.cc
baumeier: changed too

src/libxtp/gaussian_quadrature.cc
baumeier: I think this will need your intervention.

src/libxtp/gwbse/sigma_cda.cc
baumeier: I tested this for big-ish molecule: QP corrections for 25 near-gap levels, 28 threads, 3555 aux basis functions, with fixed-point updating, so in both cases 149 self-energy evaluations.

Using `.inverse()`: 6 hours 53 minutes
Using `PartialPivLU`: 6 hours 38 minutes

Linear system is about 15 minutes faster, or around 3%.
Probably doesn't hurt to keep the linear system, but the gain is a little small.

JensWehner: Yes, it is also much more stable numerically. So this was the main reason for me and it is a bit faster. If we can speed the RPA up, this will get faster too.

baumeier: OK, removed the old implementation here.

JensWehner:
It would also be nice for gaussian quadrature to have its own tests for all the grids.

JensWehner:
and the RPA needs a test as well for the new functionality.

src/tests/DataFiles/sigma_cda/c_ref.mm
```diff
@@ -0,0 +1,17 @@
...
+ 0.540951
+-0.0498701
+-0.161544
+-0.161544
+-0.342083
+-0.747893
+-0.655768
+-0.655768
+-0.389623
+-0.0743816
```
JensWehner, line 17: Wrong file format.

src/libxtp/gaussian_quadrature.cc
```diff
@@ -0,0 +1,187 @@
...
+      double num = halfpi / den;
+      Eigen::VectorXcd denominator = (DeltaE + newpoint).cwiseInverse();
+      std::complex<double> value =
+          ((Imx * (_dielinv_matrices_r[j]))
+               .cwiseProduct(denominator.asDiagonal() * Imx))
+              .sum();
+      result += num * _quadadaptedweights(j) * value;
+    }
+  }
+  return 0.5 * result.real() / (tools::conv::Pi);
```
JensWehner, line 182: Why do you take real in the end, you can use `double` all the way? What is the point behind that?

JensWehner:
@gtirimbo @baumeier 
I refactored it a bit, so that you can use the quadrature with any kind of function. 
Please write tests for it and also make sure that it gives the same results, because the current test is broken. The reference file is bullshit. 

There is still some inconsistencies in the code concerning the usage of complex data type. There was also at least one bug, because an option was not properly passed through. Please test the code seriously. 

Be consistent in your changes. This took me way longer than it was supposed to.

JensWehner:
and finally we have the CI which also builds in Debug modus to find out of bounds errors etc... if you need help with that let me know.

src/tests/DataFiles/sigma_cda/c_ref.mm
baumeier: What is the right format exactly? I see this needs the header lines for example. But is it described anywhere?

baumeier:
@JensWehner did you push changes? Which option was not passed through?

JensWehner:
I pushed now, sorry I was a bit grumpy yesterday night.

baumeier:
Any idea how to make codacy happy?

JensWehner:
die kann man ignorieren. Was noch nen bisschen fehlt ist nen Text und ggf referenzen fuer das. Fuer ppm und exact haben wir ja xavier's thesis

baumeier:
It is actually exactly what we wrote in Eqs. 28/29 of our JCP paper with the references therein.

Gaussian Quadratures are next.

gtirimbo:
> Be consistent in your changes. This took me way longer than it was supposed to.

I am really sorry I am wasting your precious time.

include/votca/xtp/sigma_base.h
```diff
@@ -44,7 +44,7 @@ class Sigma_base {
     Index rpamax = 0;
     double eta = 1e-3;
     std::string quadrature_scheme;  // Kind of Gaussian-quadrature scheme to use
-    Index order = 100;  // only needed for complex integration sigma default:12
+    Index order = 12;  // only needed for complex integration sigma default:12
```
JensWehner, line 47: actually none of the options should have a default, because defaults should be gotten from the xml files.

src/tests/test_gaussian_quadratures.cc
```diff
@@ -0,0 +1,143 @@
...
+    if (symmetry) {
+      factor = 2.0;
+    } else {
+      factor = 1.0;
+    }
+
+    return factor * exp(-std::pow(point, 2));
+  }
+};
+
```
JensWehner, line 55: modified gauss-legendre is missing from the tests

include/votca/xtp/sigma_base.h
baumeier: removed

src/tests/test_gaussian_quadratures.cc
baumeier: added

include/votca/xtp/aomatrix.h.orig
```diff
@@ -0,0 +1,106 @@
+/*
+ *            Copyright 2009-2020 The VOTCA Development Team
+ *                       (http://www.votca.org)
+ *
+ *      Licensed under the Apache License, Version 2.0 (the "License")
+ *
+ * You may not use this file except in compliance with the License.
```
JensWehner, line 7: this should not be in here.

include/votca/xtp/gw.h
```diff
@@ -59,10 +59,13 @@ class GW {
...
-    Index qp_grid_steps;     // Number of grid points
-    double qp_grid_spacing;  // Spacing of grid points in Ha
-    Index gw_mixing_order;   // mixing order
-    double gw_mixing_alpha;  //  mixing alpha, also linear mixing
+    Index qp_grid_steps;            // Number of grid points
+    double qp_grid_spacing;         // Spacing of grid points in Ha
+    Index gw_mixing_order;          // mixing order
+    double gw_mixing_alpha;         //  mixing alpha, also linear mixing
+    std::string quadrature_scheme;  // Kind of Gaussian-quadrature scheme to use
+    Index order = 100;  // only needed for complex integration sigma default:12
```
JensWehner, line 67: default and comment is misleading

src/libxtp/ImaginaryAxisIntegration.cc
```diff
@@ -0,0 +1,112 @@
...
+
+// This function calculates and stores inverses of the microscopic dielectric
+// matrix in a matrix vector
+void ImaginaryAxisIntegration::CalcDielInvVector(
+    const RPA& rpa, const Eigen::MatrixXd& kDielMxInv_zero) {
+  _dielinv_matrices_r.resize(_gq->Order());
+
+  std::cout << "\n... ... Preparing RPA for Gaussian quadrature along "
+               "imaginary axis with "
+            << _opt.order << " points" << std::endl;
```
JensWehner, line 54: Is this necessary? If so, we have to add the logger to sigma_base

src/libxtp/ImaginaryAxisIntegration.cc
```diff
@@ -0,0 +1,112 @@
...
+  boost::progress_display progress(_gq->Order(), std::cout, "... ... ",
+                                   "... ... ", "... ... ");
```
JensWehner, line 56: same here

src/libxtp/gwbse/rpa.cc
```diff
@@ -65,7 +65,7 @@ Eigen::MatrixXd RPA::calculate_epsilon(double frequency) const {
   const Index n_unocc = _rpamax - lumo + 1;
   const double freq2 = frequency * frequency;
   const double eta2 = _eta * _eta;
-#pragma omp parallel for schedule(dynamic) reduction(+ : result)
+
```
JensWehner, line 68: why is this gone?

src/libxtp/ImaginaryAxisIntegration.cc
baumeier: no

src/libxtp/ImaginaryAxisIntegration.cc
baumeier: no

src/libxtp/gwbse/rpa.cc
baumeier: No idea. Makes no sense.

include/votca/xtp/aomatrix.h.orig
baumeier: deleted

include/votca/xtp/gw.h
baumeier: default removed

src/libxtp/gwbse/rpa.cc
```diff
@@ -91,6 +92,36 @@ Eigen::MatrixXd RPA::calculate_epsilon(double frequency) const {
...
+    const Eigen::ArrayXd deltaE = _energies.tail(n_unocc).array() - qp_energy_m;
+
+    Eigen::VectorXd chi;
+    Eigen::ArrayXd deltaEm = frequency.real() - deltaE;
+    Eigen::ArrayXd deltaEp = frequency.real() + deltaE;
+
+    double sigma_1 = std::pow(frequency.imag() + _eta, 2);
+    double sigma_2 = std::pow(frequency.imag() - _eta, 2);
+
+    chi = deltaEm * (deltaEm.cwiseAbs2() + sigma_1).cwiseInverse() -
```
JensWehner, line 117: define `chi` here not above.

JensWehner (approved):
The small definition thing in the RPA, `chi` is defined way before it is used. Otherwise really nice work. Thanks to all involved. @felipeZ please review for things that are unclear.

felipeZ (approved):

src/libxtp/gwbse/rpa.cc
baumeier: done

JensWehner (approved):
Nice work

PR #551, opened 2020-09-04
Sigma derivative fix

### LINKED ISSUE #550: Sigma derivative in QPfunc with extra offset
The derivative of the self-energy in `GW::QPfunc` is returned as 

```      
return _sigma_c_func.CalcCorrelationDiagElementDerivative(_gw_level,
                                                                frequency)+
             _offset - frequency;
```

where offset is the intercept `frequency + sigma_x - Vxc` so the function effectively returns `dSigma/dw = dSigma_c/dw + Sigma_x - Vxc`

That is just a constant but shouldn't be here. Unless I understand this wrong.

baumeier:
Fixes #550

### DIFF
--- a/include/votca/xtp/gw.h
+++ b/include/votca/xtp/gw.h
@@ -111,8 +111,7 @@ class GW {
         : _gw_level(gw_level), _offset(offset), _sigma_c_func(sigma){};
     std::pair<double, double> operator()(double frequency) const {
       std::pair<double, double> value;
-      value.first = value(frequency)
-      value.second = deriv(frequency);
+      value.first = value(frequency) value.second = deriv(frequency);
       return value;
     }
     double value(double frequency) const {
@@ -121,7 +120,8 @@ class GW {
     }
     double deriv(double frequency) const {
       return _sigma_c_func.CalcCorrelationDiagElementDerivative(_gw_level,
-                                                                frequency) -1.0;
+                                                                frequency) -
+             1.0;
     }
 
    private:
--- a/include/votca/xtp/gw.h
+++ b/include/votca/xtp/gw.h
@@ -125,8 +125,7 @@ class GW {
     }
     double deriv(double frequency) const {
       return _sigma_c_func.CalcCorrelationDiagElementDerivative(_gw_level,
-                                                                frequency) +
-             _offset - frequency;
+                                                                frequency);
     }
 
    private:
--- a/include/votca/xtp/gw.h
+++ b/include/votca/xtp/gw.h
@@ -111,12 +111,8 @@ class GW {
         : _gw_level(gw_level), _offset(offset), _sigma_c_func(sigma){};
     std::pair<double, double> operator()(double frequency) const {
       std::pair<double, double> value;
-      value.first =
-          _sigma_c_func.CalcCorrelationDiagElement(_gw_level, frequency);
-      value.second = _sigma_c_func.CalcCorrelationDiagElementDerivative(
-          _gw_level, frequency);
-      value.first += (_offset - frequency);
-      value.second -= 1.0;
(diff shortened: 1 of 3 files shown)


JensWehner (approved):
Yes that was a bug. In the combined version that is not done... hmm.

JensWehner (changes requested):
actually it misses a -1.0

include/votca/xtp/gw.h
```diff
@@ -125,8 +125,7 @@ class GW {
     }
     double deriv(double frequency) const {
       return _sigma_c_func.CalcCorrelationDiagElementDerivative(_gw_level,
-                                                                frequency) +
-             _offset - frequency;
+                                                                frequency);
```
JensWehner, line 128:
```suggestion
                                                                frequency)-1.0; // -1.0 comes from derivative of the frequency itself.
```

JensWehner (approved):

include/votca/xtp/gw.h
baumeier: How? This is just the freq derivative of Sigma?

baumeier: ok, this is used in the grid solver for the root finding problem, when `f(w)=Sigma_c(w) + w0 + Sigma_x - Vxc - w = 0`

But then, the `qpweight` is calculated from this as as `1/(1 - f'(w)) = 1/(1-Sigma_c(w)' +1)`, no? That should be correcte to `1/(-f'(w))`, I guess.

JensWehner: I do not remember what the quasiparticle weight definition is.

baumeier: `1/(1-Sigma')`

JensWehner: true.

JensWehner: then you mention is correct.

include/votca/xtp/gw.h
```diff
@@ -111,7 +111,12 @@ class GW {
         : _gw_level(gw_level), _offset(offset), _sigma_c_func(sigma){};
     std::pair<double, double> operator()(double frequency) const {
       std::pair<double, double> value;
-      value.first = value(frequency) value.second = deriv(frequency);
+      value.first =
+          _sigma_c_func.CalcCorrelationDiagElement(_gw_level, frequency);
+      value.second = _sigma_c_func.CalcCorrelationDiagElementDerivative(
+          _gw_level, frequency);
+      value.first += (_offset - frequency);
+      value.second -= 1.0;
```
JensWehner: why this? The old stuff was missing a `;` but otherwise it should just call the individual functions.

JensWehner (changes requested):
why not use the `value` and `deriv` function in the `()` operator?

include/votca/xtp/gw.h
baumeier: I didn't realize that it had failed thanks to double use of `value`. Changed the return variable name.

include/votca/xtp/gw.h
```diff
@@ -110,23 +110,26 @@ class GW {
...
-      value.first += (_offset - frequency);
-      value.second -= 1.0;
-      return value;
+      std::pair<double, double> result;
+      // value.first =
+      //    _sigma_c_func.CalcCorrelationDiagElement(_gw_level, frequency);
+      // value.second = _sigma_c_func.CalcCorrelationDiagElementDerivative(
+      //    _gw_level, frequency);
+      // value.first += (_offset - frequency);
+      // value.second -= 1.0;
```
JensWehner, line 119: these comments should still go

baumeier: argh

JensWehner (approved):

Issue #552, opened 2020-09-07
Orca and %maxcore option


gtirimbo:
Sometimes it is needed to change the default `%maxcore` option in Orca. However it is not possible to change this option via the `dftgwbse.xml` input file.  

I tried something like `<orca><maxcore>3000</maxcore></orca>` with no success. 

Is it possible to fix this?

JensWehner:
it is taken from the command line parameters for your respective calculation. So `-t` on the command line.

gtirimbo:
So the `-t ` controls also the ORCA memory usage?

baumeier:
No. maxcore has nothing to do with the parallel cores/threads.

JensWehner:
ahh okay, my bad. What does the maxcore option do?

gtirimbo:
https://sites.google.com/site/orcainputlibrary/orca-common-problems

JensWehner:
I do not think that it is implemented @felipeZ ?

felipeZ:
@gtirimbo the problem is that the input option `<orca><maxcore>3000</maxcore></orca>`  is translate into
```xml
%maxcore
3000
end
```
that is not valid Orca input, right?

So, insted the `maxcore` option must be:
```xml
%maxcore 3000
```

@baumeier @gtirimbo Is there some guidelines in Orca about when a keyword must be in a single line or in multiple line?

felipeZ:
So, the rule for multi-line vs single line seems to be that if there is a single value after the keyword like, it should be in a single line,
for instance
```
%keyword value
```
Otherwise it should go into multiple line, for instance
```
%freq 
restart true
end
```

PR #553, opened 2020-09-07
fix single/multiline keywords in Orca

felipeZ:
## The problem
see #552 

## Solution 
Check whether the Orca keyword is single or multi-line and print it accordingly


JensWehner (approved):
Good work. Fixes the problem. Thanks

baumeier (approved):

PR #560, opened 2020-09-22
Dynamical screening in BSE as perturbation

baumeier:
Implements #3 by evaluating direct part of the BSE Hamiltonian at frequency of the static BSE solution and evaluating
`delta Omega_S = <S|Hd(Omega_S)-Hd(0)|S>` iteratively.

Reuses the existing BSE analysis routines as much as possible.

Still to do:
- [x] Reformulate `Analyze_IndividualContribution` to perform analysis only on a given state, not all.
- [x] tests


baumeier:
@JensWehner this is the current WIP situation. Maybe you can have a look already.

src/libxtp/gwbse/bse.cc
```diff
@@ -98,12 +100,13 @@ Eigen::MatrixXd BSE::AdjustHqpSize(const Eigen::MatrixXd& Hqp,
 }
 
 void BSE::SetupDirectInteractionOperator(
-    const Eigen::VectorXd& RPAInputEnergies) {
+    const Eigen::VectorXd& RPAInputEnergies, const double energy) {
```
JensWehner, line 103:
```suggestion
    const Eigen::VectorXd& RPAInputEnergies, double energy) {
```

src/libxtp/gwbse/bse.cc
```diff
@@ -655,5 +660,93 @@ BSE::Interaction BSE::Analyze_eh_interaction(const QMStateType& type,
...
+      /* this is a bit superfluous because we are only interested in the
+         contribution of Hd_dyn to a specific excitation, not all of them, but
+         leave for later */
+      analysis_dyn.direct_contrib =
+          Analyze_IndividualContribution(type, orb, hd_dyn);
+
+      // new energy perturbatively
+      BSEenergies_dynamic(i_exc) = BSEenergies(i_exc) +
+                                   Hd_static_contrib(i_exc) -
+                                   analysis_dyn.direct_contrib(i_exc);
```
JensWehner, line 702: This part is confusing, why should the analysis be used for perturbation theory? Why not use a a second HdOperator ?

src/libxtp/gwbse/bse.cc
```diff
@@ -655,5 +660,93 @@ BSE::Interaction BSE::Analyze_eh_interaction(const QMStateType& type,
   return analysis;
 }
 
+void BSE::Perturbative_DynamicalScreening(
+    const QMStateType& type, const Orbitals& orb,
```
JensWehner, line 664: in general the reference to the paper would be nice

src/libxtp/gwbse/bse.cc
JensWehner: I see you want to save the calculation for the static case. Considering that the saving from that will be pretty small. I would recalculate the static case here at the beginning. It does not cost much and it separates the code nicely.

JensWehner: It will cost you one iteration.

src/libxtp/gwbse/gwbse.cc
```diff
@@ -703,18 +711,39 @@ bool GWBSE::Evaluate() {
...
-      bse.Analyze_triplets(_fragments, _orbitals);
+      Hd_static_contrib_triplet = bse.Analyze_triplets(_fragments, _orbitals);
     }
 
     if (_do_bse_singlets) {
       bse.Solve_singlets(_orbitals);
       XTP_LOG(Log::error, *_pLog)
           << TimeStamp() << " Solved BSE for singlets " << flush;
-      bse.Analyze_singlets(_fragments, _orbitals);
+      Hd_static_contrib_singlet = bse.Analyze_singlets(_fragments, _orbitals);
```
JensWehner, line 730: See the comment above, I would pay the price for the one more iteration to make the code much cleaner and not have this coupling here.

src/libxtp/gwbse/bse.cc
```diff
@@ -655,5 +660,93 @@ BSE::Interaction BSE::Analyze_eh_interaction(const QMStateType& type,
...
+
+  // initial copy of static BSE energies to dynamic
+  Eigen::VectorXd BSEenergies_dynamic = BSEenergies;
+
+  // recalculate Hd at the various energies
+  for (Index i_exc = 0; i_exc < BSEenergies.size(); i_exc++) {
+    XTP_LOG(Log::info, _log) << "Dynamical Screening BSE, Excitation " << i_exc
+                             << " static " << BSEenergies(i_exc) << flush;
+
+    for (Index iter = 0; iter < _max_dyn_iter; iter++) {
```
JensWehner, line 683: Have you thought about using your anderson mixing class here as well to keep the number of iterations lower?

src/libxtp/gwbse/gwbse.cc
baumeier: Yeah, I had the static case inside the Perturbative function before, but it means that we have to recalc the static RPA. But indeed, probably no much extra effort compared to the rest of the iterations. Will go back to that one.

src/libxtp/gwbse/bse.cc
baumeier: The analysis function is used because what we need to calculate is the expectation value of Hd(dynamic) with the static BSEcoefs. And this function provides it.

baumeier: However, it calculates this for all S, not only the one we are interested in, which is a bit useless.

src/libxtp/gwbse/bse.cc
baumeier: Not really. From my, admittedly small test, but also previous experiences with the super old fortran code, this usually is done in 2-3 evaluations.

src/libxtp/gwbse/bse.cc
JensWehner: Yes, calculating the expectation value should be simply `A.T*H*A`
so I think not using the function is cleaner.

JensWehner: it might be that `H` does not provide a simple overload for matrix*vector operations only matrix*matrix

JensWehner: Factoring out a function to calculate an expectation value is probably better then using the analysis function, because the name `analysis` would be misleading.

baumeier: But `Analyze_IndividualContribution` is exactly such a function that calculates the expectation value(s). Maybe we could just rename it to avoid confusion?

JensWehner: yes, that is a good idea, but we should change the function signature then as well. Using an eigen::system and the operator should be enough. I think you also do not need the interaction struct.

src/libxtp/gwbse/gwbse.cc
baumeier: Changed this one.

src/libxtp/gwbse/bse.cc
baumeier: OK, so I have renamed the variables inside the `Perturbative_DynamicalScreening` function. The static case can actually reuse `Analyze_IndividualContribution` because that gives us all `Hd` contributions for every S at the same time, which is exactly what we need (unless we want to add an option to just do this for a selected range of states...).

Not sure what you mean by

> we should change the function signature then as well. Using an eigen::system and the operator should be enough.

baumeier:
Additional question: The `Perturbative_DynamicalScreening` function is just a `void` return, and the corrected energies are simply written to screen. Probably should be stored somewhere too, either as a separate entry in `orbitals` or overwriting the static energies. Any suggestion?

JensWehner:
Yeah that is a good question, I think you could add it to the orbitals file as a new entry. Overwriting the static bse result does not sound like a good idea, because in some sense it is a post-processing step. Storing the new energies is also not a lot of memory.

src/libxtp/gwbse/bse.cc
baumeier: > and of course the BTDA case has to be considered as well.

Indeed, and I think what `Analyze_eh_interaction` does now is maybe not correct for the full BSE, for example:

`S =    5 Omega = +13.661449286537 eV  lamdba = +90.77 nm <FT> = +22.0218 <K_x> = +0.7932 <K_d> = -9.8635`

but

`<FT> + 2<K_x> + <K_d> = 13.7447 eV` ?

include/votca/xtp/bse.h
```diff
@@ -144,9 +143,9 @@ class BSE {
   Interaction Analyze_eh_interaction(const QMStateType& type,
                                      const Orbitals& orb) const;
   template <typename BSE_OPERATOR>
-  Eigen::VectorXd Analyze_IndividualContribution(const QMStateType& type,
-                                                 const Orbitals& orb,
-                                                 const BSE_OPERATOR& H) const;
+  Eigen::VectorXd ExpectationValue_Operator(const QMStateType& type,
```
JensWehner, line 146: strictly speaking the signature should be something like `const tools::EigenSystem`,const BSE_OPERATOR& H)`

src/libxtp/orbitals.cc
```diff
@@ -601,6 +605,10 @@ void Orbitals::ReadFromCpt(CheckpointReader r) {
   r(_BSE_triplet, "BSE_triplet");
 
   r(_use_Hqp_offdiag, "use_Hqp_offdiag");
+
+  r(_BSE_singlet_energies_dynamic, "BSE_singlet_dynamic");
+
+  r(_BSE_triplet_energies_dynamic, "BSE_triplet_dynamic");
```
JensWehner, line 611: this breaks backwards compatability of the orbitals file. If you do not want that, use the version variable like here:

https://github.com/votca/xtp/blob/56d69efde879a8ffd41f1a4134b3411d7492d3e1/src/libxtp/orbitals.cc#L579

baumeier: How exactly are those versions defined now?

JensWehner: At the moment it is at version 1, so far no version 2 is used, so you can increment to version 2

JensWehner: add a comment what the version changed.

src/libxtp/gwbse/bse.cc
baumeier: Added

src/libxtp/orbitals.cc
baumeier: ok, I think I did what you suggested...

JensWehner (changes requested):
The gwbse options have to be parsed through to the appropriate xml files. Also it needs some tests. The rest looks good.

baumeier:
> The gwbse options have to be parsed through to the appropriate xml files. Also it needs some tests. The rest looks good.

Working on tests, what about the options? I still have to add them to the default xml but otherwise they are passed through, aren't they?

JensWehner:
dftgwbse.xml, iqm, eqm, qmmm

JensWehner:
The nested xml packages was not such a bad idea afterall.

baumeier:
> The nested xml packages was not such a bad idea afterall.

Yeah, for those parts that are reused in different calculators, it has some advantages...

JensWehner (approved):
Nice work

PR #570, opened 2020-10-10
Calculator refactor

JensWehner:
Removed some of the public virtual functions for calculators and used protected functions instead. 

a) reduces code duplication
b) allows us to do certain setup things for all possible calculators like e.g. libint::initialize().


JensWehner:
@votca-bot changelog: Refactored calculator interface

felipeZ (approved):
After reading [this article](http://www.gotw.ca/publications/mill18.htm) I see now the point and I agree with you

rubengerritsen (approved):
Nice, it does what you say, although I am not really sure about the libint example you mention. Because it forces us to include the libint header everywhere, but we can discuss that in the other PRs

PR #574, opened 2020-10-20
updated logger and made it cout

JensWehner:
fixes logger output.


JensWehner:
@votca-bot changelog: improved kmc output

junghans:
[boost.log](https://www.boost.org/doc/libs/1_63_0/libs/log/doc/html/index.html)?

baumeier:
That works. Needs to go to stable too.

baumeier (approved):

JensWehner:
@junghans yeah would be nice, but I still do not fully understand our logger implementation, there is some multithreading support in it. So I do not want to replace it.

JensWehner:
I guess, we could use boost log

PR #587, opened 2020-10-31
GHA: add Debug build

junghans:


JensWehner (approved):

junghans:
@certik

certik:
Cool, the Docker image seems to install in 5min. It's not actually using the Intel compiler, right? Only MKL?

https://github.com/votca/votca/runs/1337056446

junghans:
We haven't update the `votca/votca` repo yet, it uses intel e.g. here: https://github.com/votca/xtp/runs/1371280882?check_suite_focus=true

certik:
@junghans thank you! This is really useful. How do you install the trial license? I don't see it in the log.

PR #589, opened 2020-11-03
Molden file as the source for MO coefficients

rubengerritsen:
Read MO coefficients computed with orca from the molden file instead of the .gbw binary.


rubengerritsen:
@votca-bot changelog: Read MO coefficients computed with orca from the molden file instead of the .gbw binary

rubengerritsen:
I retested it to see if everything still works and it does. So if we want to switch away from the .gbw and use the molden files as the source of the MO coefficients we can merge this

src/libxtp/molden.cc
```diff
@@ -230,7 +231,12 @@ void Molden::parseMoldenFile(const std::string& filename,
       boost::trim(units);
       XTP_LOG(Log::error, _log)
           << "Reading atoms using " << units << " units." << std::flush;
-      line = readAtoms(orbitals.QMAtoms(), units, input_file);
+      // only update atoms if they are not yet loaded
+      if (orbitals.QMAtoms().size() == 0) {
+        line = readAtoms(orbitals.QMAtoms(), units, input_file);
+      } else {  // we need get the next line to continue
+        std::getline(input_file, line);
+      }
```
JensWehner, line 239: what happens if orca updated the coordinates? I do not fully understand this check

src/libxtp/qmpackages/orca.cc
```diff
@@ -312,6 +313,8 @@ bool Orca::WriteShellScript() {
   }
   shell_file << _settings.get("executable") << " " << _input_file_name << " > "
              << _log_file_name << endl;  //" 2> run.error" << endl;
+  std::string base_name = _mo_file_name.substr(0, _mo_file_name.size() - 4);
+  shell_file << "orca_2mkl " << base_name << " -molden" << endl;
```
JensWehner, line 317: should this be part of the execute script or of the parse part? Because if it is part of the parse, this change would even be backwards compatible.

rubengerritsen: Smart, I could make it part of the parse part

src/libxtp/molden.cc
rubengerritsen: Good question, what happens is that we still use the "xtp" coordinates. The reason for this if statement originally was that without it, it would just append the molden atoms to the orb object, which would mean that you get the same atoms twice in the orbitals. To prevent this I added the if statement. In principle orca shouldn't change the coordinates because that would make it different from our xtp dft implementation which seems undesirable. As an alternative I could delete the existing atoms and reread them, but I didn't see the benefit, moreover after the 8th decimal place or so we might lose a bit of accuracy by writing them to and reading them from text.

JensWehner: I do not know, before we read in the coordinates if orca uses the `optimize` keyword. I think it is better to read them in and remove the old atoms. Especially because the xyz files also do not really have 7 digits.

rubengerritsen: Okay

src/libxtp/qmpackages/orca.cc
rubengerritsen: This is more difficult than expected. There is a problem with testing the results. I can add this to the ParseMOsFile, but then I can't test the MO file parsing, because in the test we need access to a shell with orca loaded in the path. I thought to be smart and split the ParseMOsFile function in two (one with the shell script and one with the rest of the parsing) and only test the second part in the unit test, but with this virtual class stuff that can't be done, because the two separate functions aren't present in the qmpackage class. If you have any ideas on how to work around that issue I would be glad to hear. As an alternative we could not go for the backwards compatibility but raise an error if someone tries to do it the old way saying that they can convert their .gbw file with the orca_2mkl tool manually. I'll implement the latter for now.

rubengerritsen:
@votca-bot format

src/libxtp/qmpackages/orca.cc
JensWehner: Yes that is a good solution.

JensWehner (approved):

PR #593, opened 2020-11-12
center was not correctly set in AODipole

rubengerritsen:
There was an error in the setCenter function, see the change for the details.


rubengerritsen:
@votca-bot changelog: fix bug in setCenter of AODipole

JensWehner (approved):

JensWehner (changes requested):
Can you write a test, which checks this?

rubengerritsen:
Yes

include/votca/xtp/aomatrix.h
```diff
@@ -95,10 +95,12 @@ class AODipole : public AOMatrix {
 
   void setCenter(const Eigen::Vector3d& r) {
     for (Index i = 0; i < 3; i++) {
-      _r[0] = r[0];
+      _r[i] = r[i];
     }
   }  // definition of a center around which the moment should be calculated
 
+  const std::array<libint2::Shell::real_t, 3>& getCenter() const { return _r; }
```
JensWehner, line 102: I am not super happy with this because 

a) we normally do not add API just to test things
b) it exposes a second type for 3dVectors and we only want to see Eigen::Vector3d

rubengerritsen: Yeah, that is what I thought, but I was wondering what another way of testing this setCenter function would be. The only other way I could think of is setting up a full AODipole matrix with a non zero center.

JensWehner: I think overkill is better here.

JensWehner (approved):

PR #594, opened 2020-11-12
perform check in orca unit test 

rubengerritsen:
Unit test was there, but no check was done for the polar part


rubengerritsen:
@votca-bot changelog: extra check in orca unit test

JensWehner (approved):
Good catch

rubengerritsen:
That was quick

PR #599, opened 2020-11-16
AddContainer updates ID

### LINKED ISSUE #597: Wrong Qeff in fragment analysis of excitations
```
T00 ERR ...   ====== singlet energies (eV) ======
T00 ERR ...   S =    1 Omega = +3.490278561938 eV  lamdba = +355.27 nm <FT> = +6.4811 <K_x> = +0.0529 <K_d> = -3.0438
T00 ERR ...            TrDipole length gauge[e*bohr]  dx = -0.5405 dy = -0.1264 dz = +0.3415 |d|^2 = +0.4248 f = +0.0363
T00 ERR ...            HOMO-0   -> LUMO+0    : 98.4%
T00 ERR ...            Fragment    0 -- hole: 100.0%  electron:  97.6%  dQ: +0.02  Qeff: -61.44
T00 ERR ...            Fragment    1 -- hole:   0.0%  electron:   2.4%  dQ: -0.02  Qeff: +61.44
```

Calculation of the effective charge of the fragment is way off. `dQ` of the excitation seems alright, so probably the problem is somewhere in ground state contribution. Maybe related to changes with `LIBINT`.

rubengerritsen:
fixes #597 

Or at least so I think, there are two assumptions:
- That it is fine to make a copy of the incoming atoms
- That the ID's of incoming molecules always start from zero

@JensWehner are these valid assumptions?

### DIFF
--- a/include/votca/xtp/atomcontainer.h
+++ b/include/votca/xtp/atomcontainer.h
@@ -72,7 +72,7 @@ class AtomContainer {
   void AddContainer(const AtomContainer<T>& container) {
     Index offset = _atomlist.size();
     _type += "_" + container._type;
-    for (auto at : container._atomlist) {
+    for (const auto& at : container._atomlist) {
       T atom(at.getId() + offset, at.getElement(), at.getPos());
       _atomlist.push_back(atom);
     }
--- a/include/votca/xtp/atomcontainer.h
+++ b/include/votca/xtp/atomcontainer.h
@@ -70,16 +70,6 @@ class AtomContainer {
     calcPos();
   }
 
-  void AddContainer(const AtomContainer<QMAtom>& container) {
-    Index offset = _atomlist.size();
-    _type += "_" + container._type;
-    for (const auto& at : container._atomlist) {
-      QMAtom atom(at.getId() + offset, at.getElement(), at.getPos());
-      _atomlist.push_back(atom);
-    }
-    calcPos();
-  }
-
   const T& at(Index index) const { return _atomlist.at(index); }
   T& at(Index index) { return _atomlist.at(index); }
 
--- a/include/votca/xtp/atomcontainer.h
+++ b/include/votca/xtp/atomcontainer.h
@@ -70,9 +70,12 @@ class AtomContainer {
   }
 
   void AddContainer(const AtomContainer<T>& container) {
+    Index offset = _atomlist.size();
     _type += "_" + container._type;
-    _atomlist.insert(_atomlist.end(), container._atomlist.begin(),
-                     container._atomlist.end());
+    for (auto at : container._atomlist) {
+      T atom(at.getId() + offset, at.getElement(), at.getPos());
+      _atomlist.push_back(atom);
+    }
     calcPos();
   }
 
--- a/include/votca/xtp/atomcontainer.h
+++ b/include/votca/xtp/atomcontainer.h
@@ -26,6 +26,7 @@
 #include <typeinfo>
 
 // VOTCA includes
+#include "qmatom.h"
 #include <votca/tools/elements.h>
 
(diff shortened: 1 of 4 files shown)


include/votca/xtp/atomcontainer.h
```diff
@@ -70,9 +70,12 @@ class AtomContainer {
   }
 
   void AddContainer(const AtomContainer<T>& container) {
+    Index offset = _atomlist.size();
     _type += "_" + container._type;
-    _atomlist.insert(_atomlist.end(), container._atomlist.begin(),
-                     container._atomlist.end());
+    for (auto at : container._atomlist) {
```
JensWehner, line 75:
```suggestion
    for (const auto& at : container._atomlist) {
```

JensWehner: actually I am not sure about other `T` like `atom` or `staticsite`. I am not sure if this does not break other template arguements.

baumeier:
FWIW, this has fixed the problem. Not sure about the other concerns @JensWehner mentioned.

rubengerritsen:
At the moment it seems that AddContainer is only used on QMAtoms. So it should be fine. 

To make sure that it will be fine in the future as well, I propose we untemplate the `AddContainer` function and make it work for QMAtom containers only. This is a bit ugly because it breaks the general/templated usage of the `AtomContainer`, but at the same time it will give a compiler warning if someone tries to use the function. See the last commit for details.

rubengerritsen:
@votca-bot changelog: fixed atomid numbering while adding containers

rubengerritsen:
> At the moment it seems that AddContainer is only used on QMAtoms. So it should be fine.
> 
> To make sure that it will be fine in the future as well, I propose we untemplate the `AddContainer` function and make it work for QMAtom containers only. This is a bit ugly because it breaks the general/templated usage of the `AtomContainer`, but at the same time it will give a compiler warning if someone tries to use the function with another type. See the last commit for details.

I might be wrong, I am looking at the `iqm.cc` now and I don't know how the atomids of the A and B segments behave, I will try to test it out this afternoon

JensWehner:
The atomids are only needed for atomwise operations, which iqm should not do, as far as I know. 
Untemplating the add container is the right call.

rubengerritsen:
For the calculations that we do now, I agree. But the annoying thing is that the atomIDs of the B segment in iqm will always be offset by the size of the A segment if we merge the current fix in master. If someone in the future ever needs to use the atomIDs in iqm it will be very hard to figure that out. 

So I have a new idea, that I will try to implement. The QMregion in qmmm makes in essence a new "molecule" (can be multiple) that is done in the QMRegion::push_back, if we update the atom IDs there then it won't affect any other classes. Furthermore the AddContainer behaves as one should expect, by simply appending one container to another without influencing the content.

JensWehner:
Yeah yes and no, the atomid inside a molecule should be unique, even if you add two molecules to form a new one. So renumbering is actually a desirable property. For the other molecule like containers, I am not sure if the atomid can also correspond to the atomid inside the molecular dynamics simulation.

rubengerritsen:
After a QMmapping all atoms are stripped of their original (MD sim) atomIDs and hence all start at zero. My main concern raised above about the iqm was just wrong. Hence the original fix is the real fix.

JensWehner:
I thought we wanted to move this up to the QMMolecule?

rubengerritsen:
Hmm, I am not really sure that we understood one another, because that is not necessarily what I was thinking. Though it is a good idea. So hopefully I understood it correctly this time. I have added a new commit in which I moved the `AddContainer` to `QMMolecule`.

JensWehner (approved):

PR #602, opened 2020-11-26
Update xtp_autogen_mapping.in

baumeier:
GetAtomIndex call uses numpy.int64, incompatible with RDKIT's interface.


felipeZ (approved):
LGTM

PR #603, opened 2020-12-01
extract_metadata.py: fix csg parsing

junghans:


junghans:
@felipeZ @JensWehner if we use `extract_metadata.py` in csg, I would like to move it to the main repo, any objections?

felipeZ (approved):
LGTM

PR #612, opened 2020-12-09
Remove Kokkos from XTP

rubengerritsen:
The Kokkos based ewald implementation is now archived on the branch `oldKokkosEwald`. 
This PR removes the Kokkos dependent source code and the cmake parts that search for Kokkos.


rubengerritsen:
@votca-bot changelog: remove kokkos dependent ewald

junghans (approved):
Do we need to update the NOTICE file?

JensWehner (changes requested):
Thanks for the change. I think we should modify the NOTICE file and acknowledge their help when the new ewald is ready @baumeier ?

baumeier:
> Thanks for the change. I think we should modify the NOTICE file and acknowledge their help when the new ewald is ready @baumeier ?

Agreed.

JensWehner (approved):

PR #613, opened 2020-12-09
better libint lmax warning

### LINKED ISSUE #607: Libint angular momentum warning
the libint warning is a bit cryptic if the angular momentum is not high enough for a given aobasis. 

So when an AOBasis is created we should probably check for its highest L value and see if the libint version supports that and then throw an appropriate error.

I would not do it in the basisset class, because a) basissets can also be used for calculations with external dftpackages b) often a basisset contains information about chemical elements which are not used in the calculation.

rubengerritsen:
This PR adds a more explicit and helpful error message when the `lmax_limit` of libint is exceeded.

The idea is to catch the error while the basisset is normalized. Since the normalization always needs to happen, this should catch all possible `lmax_limit` exceeded errors and it avoids wrapping all `libint2::Engine` calls in a `catch-try` block.

The new error message:
```
A libint error occured:
Engine::lmax_exceeded -- angular momentum limit exceeded

You requested a computation for a shell with angular momentum 5,
but your libint package only supports angular momenta upto 4.

To fix this error you will need to reinstall libint with support
for higher angular momenta. If you installed your own libint it
should be reconfigured and installed with the option --with-max-am=<maxAngularMomentum>.
If your installed libint with your OS package manager, you will
need to setup you own libint installation with the 
--with-max-am=<maxAngularMomentum> option set.
```

fixes #607


rubengerritsen:
@votca-bot changelog: better error for exceeded angular momentum in libint

src/libxtp/aomatrices/aomatrix.cc
```diff
@@ -105,18 +105,37 @@ void AOOverlap::Fill(const AOBasis& aobasis) {
...
+        << "\nTo fix this error you will need to reinstall libint with "
+           "support\n"
+           "for higher angular momenta. If you installed your own libint it\n"
+           "should be reconfigured and installed with the option "
+           "--with-max-am=<maxAngularMomentum>.\n"
+           "If your installed libint with your OS package manager, you will\n"
+           "need to setup you own libint installation with the \n"
+           "--with-max-am=<maxAngularMomentum> option set."
+        << std::endl;
+    exit(EXIT_FAILURE);
```
JensWehner, line 137: hey great error message, but could you throw a std::runtime_error instead of `exit`. The reason is that someone might want to catch that error at some point.

JensWehner (approved):
Looking good, pls change the exit to a std::runtime_error() with the message inside. Do not `cout` it but put the messagt in the `runtime_error("message")`.

rubengerritsen:
@votca-bot format

PR #616, opened 2020-12-15
made logger easier to read

JensWehner:
The logger before relied on a pointer that was only held by osstream, which required a lot of  casts.

Now the stream has its own member streambuffer, the one problem that remains is what happens if you change the streambuffer via the public functions of ostream, but I do not see that being done in the code. Before this would have resulted in a memory leak so this is at least safer.

At least it is easier to read.


JensWehner:
@votca-bot format

junghans:
Could you also fix this new gcc warning?
https://github.com/votca/xtp/runs/1555913039

JensWehner:
I hope it fixes the warning

rubengerritsen (approved):

PR #650, opened 2021-01-18
refactored tables to not have to use dummy elements anymore

JensWehner:
Now the `setupCptTable` is a static method of each class, which separates layout and writing data better. 

@pmadhikar just if you are curious still.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -5,6 +5,7 @@ Version 2022-dev
 ================
 
 -  Updated coordinate precision (#638)
+-  Made SetupCptTable static (#650)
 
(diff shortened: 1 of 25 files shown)


JensWehner:
@votca-bot format

JensWehner:
@votca-bot changelog: Made SetupCptTable static

rubengerritsen (approved):
Nice

pmadhikar:
Good idea!

pmadhikar (commented):
Smart idea.

JensWehner:
Yeah, it is not as generalizable, but you do not have to carry the dummy argument around.

PR #662, opened 2021-01-26
density matrix for single KS state; adding single state fchk generation

baumeier:
@rubengerritsen I tried adding support for what is known as `diff2gs` density in `cubegen` to the formatted checkpoint file writer.


baumeier:
@rubengerritsen It is probably a mess though...

rubengerritsen:
@votca-bot changelog: extend fchk writer with option to print single KS state densities and densities relative to the ground state

rubengerritsen:
@votca-bot format

src/libxtp/orbitals.cc
```diff
@@ -109,6 +109,21 @@ Eigen::MatrixXd Orbitals::DensityMatrixGroundState() const {
...
+
+Eigen::MatrixXd Orbitals::DensityMatrixKSstate(const QMState& state) const {
+  if (!hasMOs()) {
+    throw std::runtime_error("Orbitals file does not contain MO coefficients");
+  }
+  if (state.Type() != QMStateType::KSstate) {
+    throw std::runtime_error("State:" + state.ToString() +
+                             " is not a Kohn Sham state");
+  }
+  Eigen::MatrixXd KSstate = _mos.eigenvectors().col(state.StateIdx());
```
JensWehner, line 122:
```suggestion
  Eigen::VectorXd KSstate = _mos.eigenvectors().col(state.StateIdx());
```

src/libxtp/orbitals.cc
```diff
@@ -109,6 +109,21 @@ Eigen::MatrixXd Orbitals::DensityMatrixGroundState() const {
   return dmatGS;
 }
 
+// Density matrix for a single KS orbital
+
+Eigen::MatrixXd Orbitals::DensityMatrixKSstate(const QMState& state) const {
+  if (!hasMOs()) {
+    throw std::runtime_error("Orbitals file does not contain MO coefficients");
+  }
+  if (state.Type() != QMStateType::KSstate) {
```
JensWehner, line 118: This should also work for PQP states.

src/libxtp/gaussianwriter.cc
```diff
@@ -67,9 +67,22 @@ std::string GaussianWriter::reorderedMOCoefficients(
...
-  Eigen::MatrixXd density = orbitals.DensityMatrixFull(state);
+
+  Eigen::MatrixXd density;
+  if (state.Type().isExciton() && diff2gs) {
+    std::array<Eigen::MatrixXd, 2> DMAT =
+        orbitals.DensityMatrixExcitedState(state);
+    density = DMAT[1] - DMAT[0];
+  } else if ((state.Type().isKSState() || state.Type().isPQPState()) &&
+             diff2gs) {
+    density = orbitals.DensityMatrixKSstate(state);
```
JensWehner, line 81: please create a function `DifferenceDensityMatrix` or so and then `if` only on `diff2gs`

JensWehner (changes requested):
Mostly fine. Small refactorings.

src/libxtp/gaussianwriter.cc
JensWehner: and a test would be helpful

rubengerritsen: Coming soon

rubengerritsen: Or not, it is covered in the test_cubefile_writer, let me know if you think that that isn't enough, but for me that is good enough

src/libxtp/orbitals.cc
```diff
@@ -109,6 +136,21 @@ Eigen::MatrixXd Orbitals::DensityMatrixGroundState() const {
   return dmatGS;
 }
 
+// Density matrix for a single KS orbital
+Eigen::MatrixXd Orbitals::DensityMatrixKSstate(const QMState& state) const {
+  if (!hasMOs()) {
+    throw std::runtime_error("Orbitals file does not contain MO coefficients");
+  }
+  if (state.Type() != QMStateType::KSstate &&
```
JensWehner, line 144: should be an ||

rubengerritsen: No, it reads a bit difficult, but it says: if it is *not* a KS state and *not* a PQP state then it will not return a KS density (and print an error). So if it is a KS state or a PQP state then it will return the KS density.

rubengerritsen:
@votca-bot format

src/libxtp/orbitals.cc
JensWehner: true

JensWehner (approved):

rubengerritsen (approved):

PR #671, opened 2021-01-28
working on new constructor

JensWehner:
Cudamatrix can now be created from a block of a matrix, which makes some things easier.

PR #672, opened 2021-01-31
Davidson refactor

JensWehner:
- Made calculation of matrix product more streamlined
- Removed lapack stuff
- added GPU support for BSE
- added another precision level to davidson
- removed full_matrix option from davidson
- removed options from xml file
merge after https://github.com/votca/xtp/pull/671

Benchmark:
Gpumuscet 28threads +2 GPUs   MKL/O3/xhost
basis 555
auxbasis 1431
full BSE 64548x64548 
old version: 495s
new version: 440s
new version+GPU: 388s

@rubengerritsen @felipeZ pls review for readability
@baumeier just have a look. :)


JensWehner:
@votca-bot changelog: removed lapack routine for solving BSE

JensWehner:
@votca-bot format

include/votca/xtp/bseoperator_btda.h
```diff
@@ -193,12 +135,13 @@ struct generic_product_impl<
...
     Map<const MatrixXd> m_reshaped(m.data(), m.rows() / 2, m.cols() * 2);
     MatrixXd temp = op._A * m_reshaped;
-    Map<MatrixXd> temp_unshaped(temp.data(), m.rows(), m.cols());
+    Map<const MatrixXd> temp_unshaped(temp.data(), m.rows(), m.cols());
     dst.topRows(half) = temp_unshaped.topRows(half);
     dst.bottomRows(half) = -temp_unshaped.bottomRows(half);
     temp = op._B * m_reshaped;
-    dst.topRows(half) += temp_unshaped.bottomRows(half);
-    dst.bottomRows(half) -= temp_unshaped.topRows(half);
+    Map<const MatrixXd> temp_unshaped2(temp.data(), m.rows(), m.cols());
```
rubengerritsen, line 142: Is this addition just for readability? Or does it also make a difference, because I thought that the Map would only contain a reference/pointer to the data.

JensWehner: yeah that is a good question. What apparently can happen is that the line `temp = op._B * m_reshaped` can lead to a move of temp. So we have to create a second map because the first now points to invalid data.

src/tests/test_bse.cc
```diff
@@ -539,6 +376,7 @@ BOOST_AUTO_TEST_CASE(bse_hamiltonian) {
     cout << Hqp_extended_ref << endl;
   }
   BOOST_CHECK_EQUAL(check_hqp_extended, true);
+  // std::cout << log;
```
rubengerritsen, line 379:
```suggestion

```
This can probably go

src/tests/test_bse.cc
```diff
@@ -335,7 +275,7 @@ BOOST_AUTO_TEST_CASE(bse_hamiltonian) {
       votca::tools::EigenIO_MatrixMarket::ReadVector(
           std::string(XTP_TEST_DATA_FOLDER) + "/bse/singlets_dynamic_full.mm");
   bool check_se_dyn_full =
-      se_dyn_full_ref.isApprox(orbitals.BSESinglets_dynamic(), 0.001);
+      se_dyn_full_ref.isApprox(orbitals.BSESinglets_dynamic(), 0.05);
```
rubengerritsen, line 278: Why has this changed so much? We now allow it to perform 50 times worse? Or is that the wrong interpretation?

rubengerritsen (approved):
Seems very nice overall. I also appreciate the effort to clean things up a bit, moving declarations to places where the variables are actually used, more error messages/handling, very nice. 

I have put a few comments here and there, but mostly questions out of curiosity.

include/votca/xtp/cudapipeline.h
```diff
@@ -58,7 +58,8 @@ class CudaPipeline {
   CudaPipeline &operator=(const CudaPipeline &) = delete;
 
   // C= A*b.asDiagonal()
-  void diag_gemm(const CudaMatrix &A, const CudaMatrix &b, CudaMatrix &C) const;
+  template <class M>
+  void diag_gemm(const M &A, const CudaMatrix &b, CudaMatrix &C) const;
```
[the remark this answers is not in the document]
JensWehner: yes, it can take `cudamatrix.transpose()` or `cudamatrix.block()` which are other classes.

include/votca/xtp/openmp_cuda.h
```diff
@@ -58,23 +71,83 @@ class OpenMP_CUDA {
...
+    }
+
+    const T& operator()() {
+      assert(p != nullptr && "Dangling reference!");
+      return *p;
+    }
+
+   private:
+    const T* p = nullptr;
+  };
```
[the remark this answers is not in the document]
JensWehner: Yes, no. Defaultable references are really evil and they sort of break any intuition about these kind of references. The only reason I use it here, is because the lifetimes are already difficult to understand and it makes the CPU and GPU code a lot more similar.

src/tests/test_bse.cc
JensWehner: good catch, this dynamic perturbation theory stuff is notoriously unstable, because it depends extremely on the convergence of the eigenvectors. 

Thinking about it this test is sort of broken, but this algoritthim is just bad, which is properly why it never caught on anywhere. 

@baumeier What do you think, should we drop this dynamic stuff again?

include/votca/xtp/cudapipeline.h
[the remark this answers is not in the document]
JensWehner: that is also the reason for the template specialization furtherdown because we have to make a copy of the `transpose()` object to pass it then as a `const T&`, it only copies the `transpose() object, which holds a reference to the underlying cudamatrix` so this does not copy the matrix but only the wrapper.

src/tests/test_bse.cc
baumeier: No.

include/votca/xtp/cudapipeline.h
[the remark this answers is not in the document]
rubengerritsen: Okay I didn't know that, thanks for the clarification.

JensWehner:
@votca-bot format

include/votca/xtp/cudapipeline.h
```diff
@@ -143,6 +149,38 @@ inline void CudaPipeline::gemm_internal(const M1 &A, const M2 &B, M3 &C,
...
 
+template <class M>
+inline void CudaPipeline::diag_gemm(const M &A, const CudaMatrix &b,
+                                    CudaMatrix &C) const {
+
+  if (b.cols() != 1 && b.rows() != 1) {
+    throw std::runtime_error("B Matrix in Cublas diag_gemm must be a vector");
+  }
+
+  cublasSideMode_t mode = CUBLAS_SIDE_RIGHT;
```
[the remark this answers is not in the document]
JensWehner: So this function multiplies a general matrix A with a diagonal Matrix D. It can either execute A*D or D*A. The mode sets the respective operation

As D is symmetric we can also write this as A^T*D or A*D. So passing in A^T uses the D*A mode.

PR #676, opened 2021-02-04
Update dipoledipoleinteraction.h

JensWehner:
Depending on the archtecture we could return a dangling reference although in practice this is typically a copy.


felipeZ (approved):
LGTM

JensWehner:
@votca-bot changelog: changed reference to a copy

PR #677, opened 2021-02-07
add CodeQL static analysis to GitHub Actions

junghans:
See https://github.com/github/codeql-action


junghans:
@jens, once this is working I think I will repackage this into a generic `code_analysis` option that is `false` by default where `codeql` can be one options, so we can e.g. `cppcheck` later.

junghans:
Ok, CodeQL needs to run for the base branch as well, similarly to coverage. So I renamed the coverage workflow into code analysis and added it in there.

@JensWehner coverage has it's own tag in the workflow, `matrix.coverage` should be we merge that into `code_analyzer = coverage`?

JensWehner (commented):
I am not really sure what codeQL does for us. We have no security threats. 

But apparently it can do lots of other things, so I am all for it.

JensWehner (approved):

PR #684, opened 2021-02-20
Davidson & Openmp fixes

### LINKED ISSUE #685: unit_test_bse unstable
From 
```
      Start 18: unit_test_bse
18/84 Test #18: unit_test_bse ...........................***Failed    0.22 sec
Running 1 test case...
A
 nan  nan  nan  nan  nan  nan  nan  nan  nan  nan
 nan  nan  nan  nan  nan  nan  nan  nan  nan  nan
 nan -nan -nan -nan  nan -nan -nan -nan -nan -nan
 nan -nan -nan -nan  nan -nan -nan -nan -nan -nan
 nan  nan  nan  nan  nan  nan  nan  nan  nan  nan
 nan  nan  nan  nan  nan  nan  nan  nan  nan  nan
 nan -nan -nan -nan  nan -nan -nan -nan -nan -nan
 nan -nan -nan -nan  nan -nan -nan -nan -nan -nan
-nan -nan -nan  nan -nan -nan -nan  nan -nan -nan
-nan -nan -nan  nan -nan -nan -nan  nan -nan -nanB

unknown location(0): fatal error: in "bse_test/bse_hamiltonian": std::runtime_error: Small generalized eigenvalue problem failed.
/__w/votca/votca/xtp/src/tests/test_bse.cc(194): last checkpoint

*** 1 failure is detected in the test module "bse_test"
```
retrigger fixes it usually.

JensWehner:
This PR does 3 things:
a) openmp_cuda functions which are supposed to be called inside a region have an explicit openthread id
b) Davidson restart no longer invalidates the last iteration
c) Davidson non-symmetric now corrects for slightly imaginary eigenvalues due to numerical noise, which lead to wrong results/crashes and other stuff before. This also allows us to use Gramschmidt orthogonalisation in all cases, which should speed up non-symmetric problems quite a bit. 

This is still a draft because I want to test it on large systems. Was tested and it works. 

Benchmark:
Gpumuscet 28threads +2 GPUs MKL/O3/xhost
basis 555
auxbasis 1431
full BSE 64548x64548
old version: 495s
new version: 440s
new version+GPU: 388s
New version+ GPU + new Davidson 310s

I wanted to use C++17 features, but it seems C++17 and intel 19 compiler and libint do not interact so well. So we are stuck with C++14. @junghans @JoshuaSBrown 

Fix #685


JensWehner:
@votca-bot changelog: redo openmp_cuda api, correct davidson for non symmetric problems.

src/libxtp/threecenter.cc
```diff
@@ -137,12 +137,15 @@ void TCMatrix_gwbse::Initialize(Index basissize, Index mmin, Index mmax,
...
-#pragma omp parallel for schedule(dynamic)
-  for (Index i = 0; i < msize(); i++) {
-    gemm.MultiplyRight(_matrix[i]);
+#pragma omp parallel
+  {
+    Index threadid = OPENMP::getThreadId();
+#pragma omp for schedule(dynamic)
+    for (Index i = 0; i < msize(); i++) {
+      gemm.MultiplyRight(_matrix[i], threadid);
+    }
```
felipeZ, line 146: @JensWehner  could you please explain this new snippet to me?  I am not very sure how the new `threadid` works

include/votca/xtp/davidsonsolver.h
```diff
@@ -114,7 +107,12 @@ class DavidsonSolver {
         storeNotConvergedData(rep, proj.root_converged, neigen);
         break;
       }
-      extendProjection(rep, proj);
+      Index extension_size = extendProjection(rep, proj);
```
felipeZ, line 110: Is this the `Index` of the last column before the extension?

JensWehner: it tells you by how much the searchspace has been extended.

src/libxtp/threecenter.cc
JensWehner: For the GPU computing you have to retireve temporary data, which depends on the threadid to avoid race conditions, I can retrieve them inside openmp_cuda, but I thought parsing it in explicitly is better, because

a) It makes it more obvious that this method should be used inside an openmp region.
b) there is a small performance benefit to not having to call OPENMP::getThreadId() each time, but instead just pass it in as a function parameter.

JensWehner: In this snippet, we first setup the threadpool with `omp parallel` and retrieve the thread id for each thread and then schedule the threads, where inside `gemm` the distribution to GPU and CPU is handled via the `threadid`

JensWehner:
@votca-bot format

src/libxtp/threecenter.cc
felipeZ: Could you add that very nice explanation somewhere in the code :) ?

include/votca/xtp/davidsonsolver.h
```diff
@@ -164,86 +159,56 @@ class DavidsonSolver {
...
       proj.AV = A * proj.V;
       proj.T = proj.V.transpose() * proj.AV;
+      if (_matrix_type == MATRIX_TYPE::HAM) {
+        proj.AAV = A * proj.AV;
+        proj.B = proj.V.transpose() * proj.AAV;
+      }
 
-    } else if (_davidson_ortho == ORTHO::GS) {
+    } else {
       /* if we use a GS ortho we do not have to recompute
```
felipeZ, line 180: could you change in the comment GS to Gramd-Schimdt

baumeier:
From the point of view of functionality, this is all very nice. I do think though that we need some developer documentation about what is going on here since it *is* pretty advanced for someone who is not *that* much of an expert. This does extend also to how the BSE operators are set up because it is very difficult to see, which call is for which part of the Hamiltonian.

baumeier (approved):
from functionality perspective

src/libxtp/gwbse/rpa.cc
```diff
@@ -83,25 +83,30 @@ Eigen::MatrixXd RPA::calculate_epsilon(double frequency) const {
...
-      denom = 4 * deltaE / (deltaE.square() + freq2);
-    } else {
-      Eigen::ArrayXd deltEf = deltaE - frequency;
-      Eigen::ArrayXd sum = deltEf / (deltEf.square() + eta2);
-      deltEf = deltaE + frequency;
-      sum += deltEf / (deltEf.square() + eta2);
-      denom = 2 * sum;
+#pragma omp parallel
+  {
+    Index threadid = OPENMP::getThreadId();
```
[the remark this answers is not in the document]
JensWehner: I am starting to disagree, the internals are all handled by the openmp_cuda class, which explains in more detail why this is used. 

I do not really want to explain it every time.

JensWehner:
Okay sure, can you tag the parts, which need more documentation? The BSE Hamiltionian?

include/votca/xtp/davidsonsolver.h
JensWehner: Do you think it needs a better name?

src/libxtp/gwbse/rpa.cc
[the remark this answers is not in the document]
JensWehner: I also argue, that the openm_cuda class was created with the intent to hide the complications of what is a happening and there is plenty of documentation in it. Whereas this routine is mostly about the physics.

include/votca/xtp/davidsonsolver.h
felipeZ: that is OK. I forgot how the algorithm works

src/libxtp/gwbse/rpa.cc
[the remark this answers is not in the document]
felipeZ: Ok :)

src/libxtp/gwbse/bse_operator.cc
```diff
@@ -58,48 +58,59 @@ Eigen::MatrixXd BSE_OPERATOR<cqp, cx, cd, cd2>::matmul(
...
-            _Mmn[v1 + vmin].block(vmin, 0, _bse_vtotal, auxsize), cd2 != 0);
+        Temp = -cd * (_Mmn[c1 + cmin].block(cmin, 0, _bse_ctotal, auxsize));
+        transform.PrepareMatrix1(Temp, threadid);
       }
       if (cd2 != 0) {
-        transform.PrepareMatrix2(
-            _Mmn[v1 + vmin].block(cmin, 0, _bse_ctotal, auxsize), cd2 != 0);
+        Temp = -cd2 * (_Mmn[c1 + cmin].block(vmin, 0, _bse_vtotal, auxsize));
+        transform.PrepareMatrix1(Temp, threadid);
       }
```
rubengerritsen, line 75: I don't really understand the logic here, if both cd and cd2 are nonzero then only the last statement has any effect and the first operation could have been left out . An if-else or if-elif-else statement would seem more appropriate in that case, the same applies to a few lines below where a similar control flow is used.

rubengerritsen (approved):
Seems fine, there is one question concerning some logic

src/libxtp/gwbse/bse_operator.cc
JensWehner: there is a `static_assert` a little higher which checks that cd and cd2 can never both be nonzero, because that makes no sense from the physics side.

JensWehner: I could use a ìf`  `else if` statement, but I wanted to keep it a bit more symmetric, but I see your point.

JensWehner: otherwise I agree I could have used an `if` `else if` case, but I thought the more symmetric thing makes more sense. I can change that, if you think it makes the code more intuitive.

felipeZ (approved):

baumeier (approved):

src/libxtp/gwbse/bse_operator.cc
rubengerritsen: Okay, I missed the static assert, that at least makes it logical, but then even an if-else would be enough and it would immediately signal that one case excludes the other. 

Change it if you like, it is a minor detail and mostly a taste thing, so if you leave it, it is also fine.

JoshuaSBrown:
> From the point of view of functionality, this is all very nice. I do think though that we need some developer documentation about what is going on here since it _is_ pretty advanced for someone who is not _that_ much of an expert. This does extend also to how the BSE operators are set up because it is very difficult to see, which call is for which part of the Hamiltonian.

Ya, what the fruitcake, was that.

PR #692, opened 2021-03-05
Problem with path of executable for orca_2mkl + problem reading the Molden file.

gtirimbo:
This should fix it.

### DIFF
--- a/src/libxtp/qmpackages/orca.cc
+++ b/src/libxtp/qmpackages/orca.cc
@@ -314,7 +314,7 @@ bool Orca::WriteShellScript() {
   shell_file << _settings.get("executable") << " " << _input_file_name << " > "
              << _log_file_name << endl;  //" 2> run.error" << endl;
   std::string base_name = _mo_file_name.substr(0, _mo_file_name.size() - 4);
-  shell_file << "orca_2mkl " << base_name << " -molden" << endl;
+  shell_file << _settings.get("executable") << "_2mkl " << base_name << " -molden" << endl;
   shell_file.close();
   return true;
 }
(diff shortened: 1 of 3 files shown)


src/libxtp/qmpackages/orca.cc
```diff
@@ -314,7 +314,7 @@ bool Orca::WriteShellScript() {
   shell_file << _settings.get("executable") << " " << _input_file_name << " > "
              << _log_file_name << endl;  //" 2> run.error" << endl;
   std::string base_name = _mo_file_name.substr(0, _mo_file_name.size() - 4);
-  shell_file << "orca_2mkl " << base_name << " -molden" << endl;
+  shell_file << _settings.get("executable") << "_2mkl " << base_name << " -molden" << endl;
```
rubengerritsen, line 317: Nice! I always just assumed it would be loaded into the path.

rubengerritsen (approved):
Nice catch!

rubengerritsen:
I see that the orca test crashes now I will have a look

baumeier:
> I see that the orca test crashes now I will have a look

Yeah, so that molden file reading error message can be a bit misleading because it throws the same "you need to use orca_2mkl" message, even if it has been done correctly but the code is looking in the wrong directory. That's what happened in QMMM and why `_run_dir + "/"` was added.
Of course, if `_run_dir` is not not properly defined, then it looks under `/` which is also meh. OTOH, the logfile reading does the same.

rubengerritsen:
I don't really agree with the misleading error message, there is a pretty clear if statement there. The reason I missed this is that for all the tools the `_run_dir` is `.`, so nothing goes wrong, but for the jobcalculators things go wrong, with these changes it should be fixed fully.

rubengerritsen:
@votca-bot changelog: fixed issue with molden file path and orca executable path

JensWehner (approved):
Thanks for the fix

baumeier (approved):

rubengerritsen:
@votca-bot format

PR #698, opened 2021-03-25
Add option for ext. field to xtp and orca

rubengerritsen:
I am trying to make the numerical polarizability calculator part of PyVotca. But I ran into problems with the options, setting an external field was very hard, because it is hidden now in the `qmpackage_defaults.xml`. What I did here is two things

1. I made the external field options part of the general package (within `dftgwbse.xml`) instead of only the xtp package
2. I added the option to print the external field to the orca input file, so one can also do calculations with an external field using orca.

My main concern with this pull request is that I understand the options handling to poorly and that there is possibly a better way of achieving the same thing. So have a critical look and let me know what you think.

### DIFF
--- a/src/libxtp/qmpackages/orca.cc
+++ b/src/libxtp/qmpackages/orca.cc
@@ -271,6 +271,28 @@ bool Orca::WriteInputFile(const Orbitals& orbitals) {
     WriteBackgroundCharges();
   }
 
+  // External Electric field
+  if (_settings.has_key("use_external_field")) {
+    if (_settings.get("use_external_field") == "true") {
+      if (_settings.has_key("externalfield")) {
+        tools::Tokenizer values(this->_settings.get("externalfield"), " ");
+        std::vector<std::string> field;
+        values.ToVector(field);
+        if(field.size() != 3){
+          throw std::runtime_error("Electric field does not have 3 values.");
+        }
+        inp_file << "%scf\n ";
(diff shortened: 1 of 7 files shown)


src/libxtp/qmpackages/orca.cc
```diff
@@ -271,6 +271,28 @@ bool Orca::WriteInputFile(const Orbitals& orbitals) {
...
+    if (_settings.get("use_external_field") == "true") {
+      if (_settings.has_key("externalfield")) {
+        tools::Tokenizer values(this->_settings.get("externalfield"), " ");
+        std::vector<std::string> field;
+        values.ToVector(field);
+        if(field.size() != 3){
+          throw std::runtime_error("Electric field does not have 3 values.");
+        }
+        inp_file << "%scf\n ";
+        inp_file << "  efield " << field[0] << ", " << field[1] << ", " << field[2] << "\n";
```
baumeier, line 285: Seems you found an undocumented Orca feature?

rubengerritsen: Yeah, but it is very googlable

JensWehner, line 285:
```suggestion
        inp_file << "  efield " << field.x() << ", " << field.y() << ", " << field.z() << "\n";
```

JensWehner (approved):
Thanks for adding this piece.

JensWehner:
@votca-bot format

JensWehner:
@votca-bot changelog: added support for external fields in DFT

PR #701, opened 2021-04-13
matrices are allocated in parallel

JensWehner:
For large systems it speeds the allocation up, the hope is also that the matrices are better distributed in 2-socket architectures

### DIFF
--- a/src/libxtp/threecenter.cc
+++ b/src/libxtp/threecenter.cc
@@ -43,7 +43,7 @@ void TCMatrix_dft::Fill(const AOBasis& auxbasis, const AOBasis& dftbasis) {
   for (Index i = 0; i < auxbasis.AOBasisSize(); i++) {
     _matrix[i] = Symmetric_Matrix(dftbasis.AOBasisSize());
   }
-  
+
   Index nthreads = OPENMP::getMaxThreads();
   std::vector<libint2::Shell> dftshells = dftbasis.GenerateLibintBasis();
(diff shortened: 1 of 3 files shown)


JensWehner:
@votca-bot changelog: allocate threecenter matrices in parallel

src/libxtp/threecenter.cc
```diff
@@ -37,8 +37,13 @@ void TCMatrix_dft::Fill(const AOBasis& auxbasis, const AOBasis& dftbasis) {
     _inv_sqrt = auxAOcoulomb.Pseudo_InvSqrt(1e-8);
     _removedfunctions = auxAOcoulomb.Removedfunctions();
   }
-  _matrix = std::vector<Symmetric_Matrix>(
-      auxbasis.AOBasisSize(), Symmetric_Matrix(dftbasis.AOBasisSize()));
+  _matrix = std::vector<Symmetric_Matrix>(auxbasis.AOBasisSize());
+
+#pragma omp parallel for schedule(dynamic, 4)
```
felipeZ, line 42: this maybe a very naive question, but I still don't get why each thread ask for 4 iterations

JensWehner: Actually it is not, I did not want the threads to lock each other up because of false sharing or so, so I thought 4 is a good number. Maybe I am wrong.

JensWehner:
@votca-bot format

felipeZ (approved):

PR #703, opened 2021-04-23
used tools::NDimVector

JensWehner:


JensWehner:
@votca-bot changelog: use NDimVector instead of std::vector in vxcgrid.

JensWehner:
@votca-bot format

felipeZ (approved):

PR #711, opened 2021-06-01
add command line option

JensWehner:
add command line option for gpus to xtp.

The implementation is not perfect, basically I use sort of a global variable, but in some sense OPENMP does the same, so maybe it is okay at least for now.

### DIFF
--- a/src/libxtp/openmp_cuda.cc
+++ b/src/libxtp/openmp_cuda.cc
@@ -23,6 +23,28 @@
 namespace votca {
 namespace xtp {
 
+// Has to be declared because of
+// https://stackoverflow.com/questions/9110487/undefined-reference-to-a-static-member
+Index OpenMP_CUDA::number_of_gpus=0;
+
+Index OpenMP_CUDA::UsingGPUs() { return number_of_gpus; }
+
+Index OpenMP_CUDA::AvailableGPUs() {
+#ifdef USE_CUDA
+  return count_available_gpus();
+#else
+  return 0;
+#endif
+}
+
+void OpenMP_CUDA::SetNoGPUs(Index number) {
+  if (number < 0 || number > AvailableGPUs()) {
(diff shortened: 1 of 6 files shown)


JensWehner:
@votca-bot format

JensWehner:
@votca-bot changelog: add command line option for number of gpus

src/libxtp/openmp_cuda.cc
```diff
@@ -465,7 +487,8 @@ Eigen::MatrixXd OpenMP_CUDA::getReductionVar() {
   for (Index i = 0; i < Index(gpus_.size()); i++) {
     GPU_data& gpu = gpus_[i];
     gpu.activateGPU();
-    cpus_[i].reduce() = *(gpu.temp.back());
+    Eigen::MatrixXd temp=*(gpu.temp.back());
```
felipeZ, line 490: Out of curiosity why do you need this temp array here?

felipeZ (approved):

src/libxtp/openmp_cuda.cc
JensWehner: Good point I do not :)

src/libxtp/openmp_cuda.cc
```diff
@@ -465,7 +487,8 @@ Eigen::MatrixXd OpenMP_CUDA::getReductionVar() {
   for (Index i = 0; i < Index(gpus_.size()); i++) {
     GPU_data& gpu = gpus_[i];
     gpu.activateGPU();
-    cpus_[i].reduce() = *(gpu.temp.back());
+    Eigen::MatrixXd temp = *(gpu.temp.back());
+    cpus_[i].reduce() += temp;
```
JensWehner, line 491:
```suggestion
    cpus_[i].reduce() = *(gpu.temp.back());
```

PR #722, opened 2021-07-12
Update qmm moptions

rubengerritsen:
While preparing the tutorial we found it hard to do a QMMM calculation on a few or just one segment. I have added an option in the `qmmm` to take care of that.

We also need an option to use ground state geometries for the excited state calculations. Now for an `s1` state calculation the job file will have
```
<segments>0:s1</segments>
```
we also want it to be able to print `<segments>0:n</segments>` for the `s1` state. This option will be added shortly

### DIFF
--- a/share/xtp/xml/qmmm.xml
+++ b/share/xtp/xml/qmmm.xml
@@ -7,7 +7,7 @@
     <job_file help="name of jobfile to which jobs are written" default="qmmm_jobs.xml"/>
     <write_parse>
       <states help="states to write jobs for and which to parse from jobfile" default="n e h"/>
-      <segments help="segments to write jobs for and which to parse from jobfile" default="all" choices="all, index, list of segment indices"/>
+      <segments help="segments to write jobs for and which to parse from jobfile put either the word all or a list of numbers i.e. 0 4 5 etc." default="all"/>
     </write_parse>
   </qmmm>
 </options>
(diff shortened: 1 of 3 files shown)


JensWehner:
Sorry what, I do not understand

JensWehner:
I am not sure that is needed.

rubengerritsen:
Yeah it is, if you want to do QMMM on a single segment you can't (or I don't know how). The example use case for the tutorial is acetone in water, we don't want to do QM on water. We only want to do the QM on the acetone but in a polarizable environment. Now you first create the complete job file with `xtp_parallel blablabla`, you get a really long job file, next delete all the jobs you don't need (you only need one or two so you delete approx 99.9% of all jobs manually) and change the geometry back to `n` manually. Now you can do that stuff in the options.

JensWehner:
Just write a small jobfile by hand?

JensWehner:
Maybe your idea is really good actually, let me have a think.

JensWehner:
So I had a thought. I agree that it is a good idea to restrict the number of segments jobs are created for. 

For the second part I am not so sure about, because in the end the jobfile has a lot of flexibility and there is no way we can produce all of it from the calculator options. I always expected users to either use python or manually write the complicated jobfiles. 

For the tutorial I would just say write the 1-3 jobs you want to perform per hand and put it in the tutorial. We also do that in the other tutorials. 

If you want to create more complicated jobfiles from the options, I would like to have a better syntax. 

So to summarize for the tutorial I do not see the need, should we provide more python scripts, like https://github.com/votca/xtp/pull/535 or have a terse syntax to make jobfile creation more flexible( I would like to have that for eqm and stuff as well then)

PR #723, opened 2021-07-13
fixes to options refactor

rubengerritsen:
I will collect all minor things that might need changing with suggested changes in this PR.


JensWehner:
Thanks I missed those things.

PR #724, opened 2021-07-14
add qmmm options

rubengerritsen:
These are the same extra options as in the previous PR but now based on the `options_refactor` branch.

### DIFF
--- a/share/xtp/xml/qmmm.xml
+++ b/share/xtp/xml/qmmm.xml
@@ -5,9 +5,11 @@
     <max_iterations help="max iterations for qmmm scf loop" default="100" choices="int+" />
     <map_file help="xml file with segment definition" default="votca_map.xml" />
     <job_file help="name of jobfile to which jobs are written" default="qmmm_jobs.xml" />
-    <io_states help="states to write jobs for and which to parse from jobfile" default="n e h" />
+<io_jobfile help="options to write a jobfile">
+    <states help="states to write jobs for and which to parse from jobfile" default="n e h" />
     <segments help="segments to write jobs for and which to parse from jobfile" default="all"/>
     <use_gs_for_ex help="If true uses the ground state geometry for excited states" default="false"/>
+  </io_jobfile>
     <regions list="" help="Definitions of all regions inside qmmmm" default="REQUIRED">
       <qmregion default="OPTIONAL" link="region.xml" help="qmregion with dft and optionally gwbse">
         <state help="initial state to track or ground state for dftcalculations" default="n" />
--- a/share/xtp/xml/qmmm.xml
+++ b/share/xtp/xml/qmmm.xml
@@ -5,11 +5,11 @@
     <max_iterations help="max iterations for qmmm scf loop" default="100" choices="int+" />
     <map_file help="xml file with segment definition" default="votca_map.xml" />
     <job_file help="name of jobfile to which jobs are written" default="qmmm_jobs.xml" />
-<io_jobfile help="options to write a jobfile">
-    <states help="states to write jobs for and which to parse from jobfile" default="n e h" />
-    <segments help="segments to write jobs for and which to parse from jobfile" default="all"/>
-    <use_gs_for_ex help="If true uses the ground state geometry for excited states" default="false"/>
-  </io_jobfile>
+    <io_jobfile help="options to write a jobfile">
+      <states help="states to write jobs for and which to parse from jobfile" default="n e h" />
+      <segments help="segments to write jobs for and which to parse from jobfile" default="all"/>
+      <use_gs_for_ex help="If true uses the ground state geometry for excited states" default="false"/>
+    </io_jobfile>
     <regions list="" help="Definitions of all regions inside qmmmm" default="REQUIRED">
       <qmregion default="OPTIONAL" link="region.xml" help="qmregion with dft and optionally gwbse">
         <state help="initial state to track or ground state for dftcalculations" default="n" />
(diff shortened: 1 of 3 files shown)


src/libxtp/jobcalculators/qmmm.cc
```diff
@@ -45,8 +46,21 @@ void QMMM::ParseSpecificOptions(const tools::Property& options) {
...
+  whichSegments_ = options.get(".segments").as<std::string>();
+
+  if (whichSegments_ == "all") {
+    all_segments_ = true;
+  } else {
+    all_segments_ = false;
+    std::stringstream ss(whichSegments_);
+    Index segID;
+    while (ss >> segID) {
+      segments_.push_back(segID);
```
JensWehner, line 61: Can you use the indexparser class for this it allows you also to specify ranges of indeces e.g. `1 3:9 13` and will convert that to a nice list of indeces.

src/libxtp/jobcalculators/qmmm.cc
```diff
@@ -225,36 +239,55 @@ void QMMM::WriteJobFile(const Topology& top) {
...
-      job.ToStream(ofs);
-      jobid++;
     }
   }
 
   ofs << "</jobs>" << std::endl;
   ofs.close();
   std::cout << std::endl
-            << "... ... In total " << jobid + 1 << " jobs" << std::flush;
+            << "... ... In total " << jobid << " jobs" << std::flush;
```
JensWehner, line 264: why did you remove the `+1` ?

src/libxtp/jobcalculators/qmmm.cc
```diff
@@ -225,36 +239,55 @@ void QMMM::WriteJobFile(const Topology& top) {
...
+        jobid++;
+      }
+    }
+  } else {
+    for (Index segID : segments_) {
+      const Segment& seg = top.Segments()[segID];
+      for (const QMState& state : states_) {
+        Job job = createJob(seg, state, jobid);
+        job.ToStream(ofs);
+        jobid++;
```
JensWehner, line 256: I would prefer we filled the segments_ vector completely for `all` at the beginning so we do not have to carry around the `all_segments_` bool and do not have to have an if here at all.

src/libxtp/jobcalculators/qmmm.cc
```diff
@@ -225,36 +239,55 @@ void QMMM::WriteJobFile(const Topology& top) {
...
+  pInput.add("site_energies", marker);
+  tools::Property& regions = pInput.add("regions", "");
+  tools::Property& region = regions.add(getFirstRegionName(), "");
+  region.add("id", "0");
+  if (hasQMRegion()) {
+    region.add("state", state.ToString());
+  }
+  if (use_gs_for_ex_ && (state.Type() == QMStateType::Singlet ||
+                         state.Type() == QMStateType::Triplet)) {
+    region.add("segments", std::to_string(seg.getId()) + ":n");
```
JensWehner, line 283: This is a nice refactor thx. :)

src/libxtp/jobcalculators/qmmm.h
```diff
@@ -47,13 +47,18 @@ class QMMM final : public ParallelXJobCalc<std::vector<Job> > {
...
 
   std::pair<std::string, tools::Property> regions_def_;
 
   Index max_iterations_;
   bool print_regions_pdb_ = false;
+  bool use_gs_for_ex_ = false;
   std::vector<QMState> states_;
+  std::string whichSegments_;
+  std::vector<Index> segments_;
+  bool all_segments_ = true;
```
JensWehner, line 61:
```suggestion
  std::vector<Index> segments_to_write_;
```
should be enough

src/libxtp/jobcalculators/qmmm.h
```diff
@@ -47,13 +47,18 @@ class QMMM final : public ParallelXJobCalc<std::vector<Job> > {
 
  private:
   bool hasQMRegion() const;
+  Job createJob(const Segment& seg, const QMState& state, Index jobid);
```
JensWehner, line 50:
```suggestion
  Job createJob(const Segment& seg, const QMState& state, Index jobid)const;
```
Could we make it `const` ?

JensWehner (commented):
I am not super happy with the bool but it is fine the other idea is really good

share/xtp/xml/qmmm.xml
```diff
@@ -6,6 +6,8 @@
     <map_file help="xml file with segment definition" default="votca_map.xml" />
     <job_file help="name of jobfile to which jobs are written" default="qmmm_jobs.xml" />
     <io_states help="states to write jobs for and which to parse from jobfile" default="n e h" />
+    <segments help="segments to write jobs for and which to parse from jobfile" default="all"/>
+    <use_gs_for_ex help="If true uses the ground state geometry for excited states" default="false"/>
```
JensWehner, line 10:
```suggestion
<io_jobfile help="options to write a jobfile">
    <states help="states to write jobs for and which to parse from jobfile" default="n e h" />
    <segments help="segments to write jobs for and which to parse from jobfile" default="all"/>
    <use_gs_for_ex help="If true uses the ground state geometry for excited states" default="false"/>
  </io_jobfile>
```
I think this would be clearer. What do you think?

src/libxtp/jobcalculators/qmmm.cc
rubengerritsen: because it is wrong, at the end of the for loop above you end with `jobid++` so it always prints one job to much in the summary

JensWehner: good point. :)

rubengerritsen:
In the `necessary evil` commit, we need to keep the string of the segments around until we are in the `WriteJobFile` function, because we don't have access to the number of segments yet in the parse options function.

JensWehner:
@votca-bot format

JensWehner (approved):

PR #725, opened 2021-07-15
Create GPU.yml

JensWehner:


JensWehner:
@votca-bot changelog: added basis gpu runner and test to suite

.github/workflows/GPU.yml
```diff
@@ -0,0 +1,43 @@
...
+    branches:
+      - master
+
+jobs:
+  build:
+    runs-on: [self-hosted, GPU, benchmark]
+
+    steps:
+      - name: cleanup
+        run: rm -rf *
```
junghans, line 14: Is that actually needed?

.github/workflows/GPU.yml
```diff
@@ -0,0 +1,43 @@
+name: GPU
+
+on:
+  pull_request:
+    branches:
+      - master
```
junghans, line 6: I guess ultimately we want it on stable too.

.github/workflows/GPU.yml
```diff
@@ -0,0 +1,43 @@
...
+      - name: Checkout xtp
+        uses: actions/checkout@v2
+        with:
+          path: votca/xtp
+          
+      - name: Configure Cmake
+        run: cmake -B builddir -DBUILD_CSGAPPS=ON -DBUILD_XTP=ON -DENABLE_TESTING=ON -DCMAKE_BUILD_TYPE=Release -DUSE_CUDA=ON votca
+        
+      - name: Build
+        run: cmake --build builddir --parallel 6
```
junghans, line 32: Where is the 6 coming from?

.github/workflows/GPU.yml
```diff
@@ -0,0 +1,43 @@
...
+        
+      - name: run Tests
+        run: |
+              cd builddir
+              ctest
```
junghans, line 37: `ctest --output-on-failure`

.github/workflows/GPU.yml
```diff
@@ -0,0 +1,43 @@
...
+          
+      - name: Configure Cmake
+        run: cmake -B builddir -DBUILD_CSGAPPS=ON -DBUILD_XTP=ON -DENABLE_TESTING=ON -DCMAKE_BUILD_TYPE=Release -DUSE_CUDA=ON votca
+        
+      - name: Build
+        run: cmake --build builddir --parallel 6
+        
+      - name: run Tests
+        run: |
+              cd builddir
```
junghans, line 36: Use ‘`working-directory:` property instead.

junghans (changes requested):

.github/workflows/GPU.yml
JensWehner: I thought so, or how do I make sure the environment does not persist?

.github/workflows/GPU.yml
JensWehner: yes, after the next release.

.github/workflows/GPU.yml
JensWehner: The machine it is running on has 6 cores

JensWehner:
@junghans concerning caching what do you think would be best? ccache or some GH Actions package?

.github/workflows/GPU.yml
JensWehner: If I do not do it, it checks out into a dir that already exists and throws an error

JensWehner:
On a related node, I would like to run this https://github.com/votca/xtp/blob/master/share/benchmark/xtp_benchmark.in
afterwards, but I am not exactly sure how to source votca, should I put the script into cmake?

.github/workflows/GPU.yml
junghans: I would just use `$(nproc)` here.

.github/workflows/GPU.yml
```diff
@@ -0,0 +1,41 @@
...
+          path: votca/xtp
+          
+      - name: Configure Cmake
+        run: cmake -B builddir -DBUILD_CSGAPPS=ON -DBUILD_XTP=ON -DENABLE_TESTING=ON -DCMAKE_BUILD_TYPE=Release -DUSE_CUDA=ON votca
+        
+      - name: Build
+        run: cmake --build builddir --parallel 6
+        
+      - name: run Tests
+        run: ctest --output-on-failure
```
junghans, line 35: We should run the rest in parallel too.

.github/workflows/GPU.yml
```diff
@@ -0,0 +1,41 @@
...
+
+    steps:
+      - name: cleanup
+        run: rm -rf *
+      - name: Checkout votca repo
+        uses: actions/checkout@v2
+        with:
+          repository: votca/votca
+          submodules: true
+          ref: master
```
junghans, line 20: When we ultimately want enable it for stable, that should be the target branch, so something like `${{ github.base_ref }}` for pull request and `${{ github.ref }} for branch builds.

.github/workflows/GPU.yml
junghans: You should add `stable` here then, it won’t trigger until the workflow is on that branch, but at least then we don’t forget.

.github/workflows/GPU.yml
JensWehner: how can I do either?

.github/workflows/GPU.yml
```diff
@@ -0,0 +1,42 @@
...
+
+    steps:
+      - name: cleanup
+        run: rm -rf *
+      - name: Checkout votca repo
+        uses: actions/checkout@v2
+        with:
+          repository: votca/votca
+          submodules: true
+          ref: ${{ github.base_ref }}
```
junghans, line 21: Problem with that is: "This property is only available when the event that triggers a workflow run is either pull_request or pull_request_target.", so it will fail when triggered by a push to master.

I think you will need to a little magic like in the prepare step:
```bash
if [[ ${GITHUB_REF} = refs/pull/*/merge ]]; then # pull request
  ref="${GITHUB_BASE_REF}"
elif [[ ${GITHUB_REF} = refs/heads/stable ]]; then # stable branch
  ref="stable"
else
  ref="master"
fi
echo "::set-output name=ref::${ref}"
```

And the use `${{ steps.prepare.outputs.ref }}` here.

.github/workflows/GPU.yml
```diff
@@ -0,0 +1,42 @@
...
+  pull_request:
+    branches:
+      - master
+      - stable
+
+jobs:
+  build:
+    runs-on: [self-hosted, GPU, benchmark]
+
+    steps:
```
junghans, line 13: add `id: prepare` here. So can use `steps.prepare.outputs.ref` later.

junghans:
Also once we are done, we should squash all the commits first!

.github/workflows/GPU.yml
junghans: See my comment below.

junghans (approved):

PR #737, opened 2021-09-14
made xtp ready for c++17

JensWehner:
I had to tell libint to not use constexpr arrays and then the old intel compilers work. 
I could do this dependent on the config but it was not worth the hassle. 

To make this work I place all major libint calls into one .cc file. Should also reduce compile time, because the headers do not have to be parsed as often. 

This might still fail with a memory leak in tools


JensWehner:
@votca-bot format

JensWehner:
@votca-bot changelog: upgrade xtp to cxx17

rubengerritsen (approved):

PR #741, opened 2021-09-24
Ewald

rubengerritsen:
This is the basis for the ewald implementation. There are still some things to do:

- The final energy evaluation is still wrong:
    - For the k-space part it is simply not implemented,
    - For the real-space part it results in NaN at the moment.
- Unit tests are still missing
- The parallelization is tricky, parts were parallelized at some point, but with unpredictable results as a consequence. 

The plan is to add unit tests until everything is covered. I also expect that creating the unit tests will help to resolve the energy evaluation issues that are there currently.


JensWehner:
@rubengerritsen do you want a review already?

rubengerritsen:
No, a bit to early, let me fix the unit tests and energy evaluation first, then I will press the button for a review. I just made this because I promised to move it from my fork to here last week.

junghans:
Replaced by https://github.com/votca/votca/pull/847

PR #743, opened 2021-09-27
Fix for the hdf5 issue

rubengerritsen:
This PR probably fixes the issues with the HDF5 on FreeBSD. 

We tested it on Mac OS and there the changes solve the issue. We changed from using the `std::string` writer to write "C strings". There is also an extra test in the aobasis, that should pick up differences between written strings and the ones read back in.

### DIFF
--- a/include/votca/xtp/checkpointreader.h
+++ b/include/votca/xtp/checkpointreader.h
@@ -153,11 +153,11 @@ class CheckpointReader {
   const std::string path_;
   void ReadScalar(const CptLoc& loc, std::string& var,
                   const std::string& name) const {
-    const H5::DataType* strType = InferDataType<std::string>::get();
 
     H5::Attribute attr = loc.openAttribute(name);
+    H5::StrType stype = attr.getStrType();
 
-    attr.read(*strType, var);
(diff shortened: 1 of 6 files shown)


rubengerritsen:
@votca-bot format

rubengerritsen:
@votca-bot changelog: fix issue with writing strings to hdf5 files

JensWehner (approved):
Thanks for the fix. Looking good

include/votca/xtp/checkpointreader.h
```diff
@@ -153,15 +153,11 @@ class CheckpointReader {
...
 
     H5::Attribute attr = loc.openAttribute(name);
+    H5::StrType stype = attr.getStrType();
 
-    H5std_string readbuf("");
-
-    attr.read(*strType, readbuf);
-
-    var = readbuf;
+    attr.read(stype, var);
```
JensWehner, line 160: are we sure we do not actually have to resize it before passing it to read?

junghans:
Replaced by https://github.com/votca/votca/pull/848

